PR #55, opened 2026-09-21
House rules: local gate and e2e, PR review on every PR, agent merges, legacy fence

ericeslinger:
## What

Sets the working process for the rewrite, per the discussion on 2026-09-21:

- **No CI.** `gate.yml` removed; `release.yml` is manual (`workflow_dispatch`). The CI's tarball install-and-play step is now `npm run e2e` (`scripts/e2e.sh`).
- **`CLAUDE.md`** (new): gate before every commit, e2e before a PR, receipts in the PR body; all work on a branch and merged by PR; the `pr-review` agent on every PR; **the coding agent may merge its own PR** when the review is clean, until v1.0; the legacy fence.
- **The legacy fence:** `vitest` excludes `**/legacy/**` in both workspaces; `corpus/legacy/` exists with a README; `check-corpus` already reads only `good`/`bad`. Nothing is moved yet — the gate is green today, and each backlog PR fences what it breaks.
- **`.claude/agents/pr-review.md`** rewritten for this repo. It was a verbatim copy of overstory's (callables, seeders, `gh pr checks`); it now reviews against `CLAUDE.md`, the spec sections the issue names, and the working notes' *Decisions*, and **re-runs the gate at the PR head in a worktree**, since it is the only check that runs anywhere but the author's machine.
- README's Developing section updated.

## Why

GitHub minutes are not to be spent on this repository until it is stable, and the rewrite will break most existing suites correctly; the fence keeps "green before push" a rule that can actually be kept.

## Receipts

- gate at `e09df9a`: green — `✓ corpus: every good archive passes, every bad one fails as expected`
- e2e at `e09df9a`: green — `e2e: green (pottery-studio installed from tarballs and played at e09df9a)`

Not a backlog item: process, not language. Fenced: nothing.

https://claude.ai/code/session_01VmTaQx9p8keChvVVw6v3Ro


ericeslinger:
**Non-blocking** — `scripts/e2e.sh:20`

The trailing `$(git -C "$(dirname "$0")/.." rev-parse --short HEAD 2>/dev/null || echo HEAD)` never resolves to the actual commit; it always falls back to the literal string `HEAD`.

Reproduced by running `npm run e2e` at `f5b8654` in a clean worktree: the last line prints `e2e: green (pottery-studio installed from tarballs and played at HEAD)`. The script's earlier `cd "$(dirname "$0")/.."` correctly resolves once, from the original cwd, but by the time the final `echo` runs the script has done `cd / && rm -rf "$sandbox" "$packs"`; `$0` is still the relative path `scripts/e2e.sh`, so `dirname "$0"` (`scripts`) resolved against cwd `/` is `/scripts`, which doesn't exist, and `git -C` fails silently into the `HEAD` fallback every time. Harmless today since the PR body's e2e receipt is hand-written rather than pasted verbatim, but the script itself can never produce the sha CLAUDE.md's "receipts" rule expects it to carry. Fix: capture the repo root once at the top (`root=$(cd "$(dirname "$0")/.." && pwd)`) and use `git -C "$root"` at the end.

ericeslinger:
**Non-blocking** — `.claude/agents/pr-review.md:74-77`

"Read, in this order" step 4 lists "prose audiences" as one of the areas with a matching *Decisions* entry in `sprout-working-notes.md`, but no such decision exists there.

Checked `docs/design/sprout-working-notes.md`'s *Decisions* section (grep for "audience" returns nothing): it covers type parameters, roles, stateful text, weighted variation, disambiguation, stored references, nickname expiry, the role cap, the step budget, actor-less turns, silence, typing and the poll — no entry about who a passage is written for. The spec does have a `### The three audiences` section, so a reviewer sent to check "the matching Decisions" for a prose-audience change will find nothing to match against and may either invent a check or skip it. Either drop "prose audiences" from the list, or add the missing decision to the working notes if one was actually made.

ericeslinger:
**Non-blocking** — `scripts/check-corpus.mjs` (unchanged by this PR) vs. `CLAUDE.md`'s legacy fence

The legacy-fence rule names `sprout/examples/legacy/…` as a valid fence location, but `check-corpus.mjs` hardcodes `EXAMPLES = ['pottery-studio', 'wanderers-shed']` under `sprout/examples/`, so fencing either example would crash `npm run check` (and therefore `npm run gate`) rather than fail it cleanly.

Verified: `node cli/bin/sprout.js check sprout/examples/does-not-exist --json` prints a plain-text error to stderr and no JSON to stdout; `check-corpus.mjs`'s catch block does `JSON.parse(String(err.stdout))` unconditionally, which throws `Unexpected end of JSON input` on empty stdout — an uncaught exception, not a clean red gate. Nothing here breaks today (this PR fences nothing), but the fence rule's bullet list of what a fencing PR must do ("moved... named in the PR, with the reason and the issue that will delete it") doesn't mention that fencing an example also requires editing `check-corpus.mjs`'s hardcoded list, and the failure mode when someone forgets is an unhelpful stack trace rather than a legible gate failure. Worth a line in the fence rule, or teaching the script to scan the directory instead of naming examples.

ericeslinger:
**Non-blocking** — `README.md`, `## Design`

The Design section wasn't touched by this PR and now disagrees with the orientation `CLAUDE.md` sets up in the same diff.

`CLAUDE.md` tells a reader that `sprout-design-spec.md` is where the language is authoritative, with `sprout-working-notes.md` and `sprout-build-backlog.md` alongside it, and that "older files under `docs/design/` are dated and superseded." README's `## Design` section, three paragraphs below the `## Developing` section this PR did update, still only points a new reader at `sprout.md`, `2026-09-17-sprout-split.md` and `sprout-worked-example/` — the very files CLAUDE.md now calls superseded — and never mentions the spec, the working notes, the backlog or the 2026-09-20 reviews. Not this PR's stated scope, but it's the PR that made the omission visible.

ericeslinger:
**Review**: no blocking findings, 4 non-blocking. Gate at `f5b8654`: green (reproduced in a fresh worktree — lint, format, both builds, 356 tests, typecheck:spec, `sprout check` over corpus and examples). `npm run e2e` also reproduced green. Both match the PR body's receipts.

This is a process PR with no code to run against the design spec, so I checked it for internal consistency instead: CLAUDE.md's section names, the backlog's 53 items and issue numbers, and the spec sections `pr-review.md` cites all check out exactly against the PR head and the live issue tracker. The rewritten `pr-review.md`'s own instructions (pin to FETCH_HEAD, worktree gate re-run, the literal script) all worked as written. The four non-blocking findings above are: `scripts/e2e.sh`'s sha in its own last line never resolves (always prints literal `HEAD`); `pr-review.md` cites a "prose audiences" Decision that doesn't exist in the working notes; `check-corpus.mjs`'s hardcoded example list would crash rather than fail cleanly the first time an example is fenced, which the fence rule doesn't call out; and README's `## Design` section wasn't updated to match the new orientation CLAUDE.md sets up in this same PR.

ericeslinger:
All four findings fixed in e09df9a, each as described:

1. `scripts/e2e.sh` — the sha is now taken at the top, before the script leaves the repo; the receipt line prints the real short sha (`e2e: green (… played at e09df9a)`).
2. `.claude/agents/pr-review.md` — the *Decisions* list now names the entries that exist in the working notes: actor-less turns and the prose audiences (one decision), silence, typing, the poll, stored references, the limits' numbers.
3. `scripts/check-corpus.mjs` — examples are discovered as directories under `sprout/examples/` holding a `sprout.json`, so one moved under `legacy/` is simply not an example; no hard-coded names, no uncaught exception.
4. `README.md` — the Design section now leads with the spec, the working notes, the backlog and the reviews, and names the dated files as the previous language kept as history.

Receipts at `e09df9a`: gate green (`✓ corpus: every good archive passes, every bad one fails as expected`); e2e green (`e2e: green (pottery-studio installed from tarballs and played at e09df9a)`).

PR #56, opened 2026-09-21
Phase 0 — foundations: spans, limits, the closed bundle, strict and lenient

### LINKED ISSUE #1: B01 · Position spans on every AST node, so every diagnostic names a line and column
**Backlog item B01** · Phase 0 — Foundations · size M

## What

Position spans on every AST node, so every diagnostic names a line and column

## Spec

Implements: **The compiler › Diagnostics** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

Nothing — this can be started now.

## Blocks

#3 (B03), #50 (B50)

## Working on this

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B01.

### LINKED ISSUE #2: B02 · Runtime budgets as host configuration with the spec's defaults
**Backlog item B02** · Phase 0 — Foundations · size M

## What

Runtime budgets as host configuration with the spec's defaults: steps (parsing, range walks and `each` charged), output per recipient, events, cascade depth, passage depth, spawns per turn, live instances, wake floor, poll steps, wall-clock backstop

## Spec

Implements: **Limits** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

Nothing — this can be started now.

## Blocks

#15 (B15), #39 (B39), #43 (B43)

## Working on this

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B02.

### LINKED ISSUE #3: B03 · Closed-bundle compile
**Backlog item B03** · Phase 0 — Foundations · size L

## What

Closed-bundle compile: vendored libraries, content hashing, whole-bundle checking, and the bundle artifact carrying the word set, level, pinned extensions, library hashes and the caps checked against

## Spec

Implements: **Kinds › Libraries and namespaces; The compiler › Two tiers, What compiling produces** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#1 (B01)

## Blocks

#4 (B04), #5 (B05), #8 (B08), #12 (B12), #19 (B19), #21 (B21), #43 (B43)

## Working on this

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B03.

### LINKED ISSUE #4: B04 · Strict publish and lenient load
**Backlog item B04** · Phase 0 — Foundations · size M

## What

Strict publish and lenient load: any problem refuses at publish; at load a broken or withheld file reads as absent with exactly the behaviour in the absent table, and stored state for absent objects is kept

## Spec

Implements: **The compiler › Strict and lenient, What absent means** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#3 (B03)

## Blocks

Nothing.

## Working on this

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B04.

ericeslinger:
All four of Phase 0, one commit each with the gate green at every one, plus two commits carrying Eric's decision on the manifest and the review's findings.

Closes #1
Closes #2
Closes #3
Closes #4

## What, and why

**B01 · Position spans on every AST node** — `source.ts`, `nodes.ts`, `diagnostics.ts`, `lexer.ts`.

Every diagnostic has to name the line and column of the token it is about, and it can only do that if the node it complains about remembers which token it came from. That is a property of the data model rather than of the error strings. A span is two offsets and a file reference, because every node carries one and almost none compute from one; line and column come from a lazily built index when something is actually reported. The rule is made mechanical rather than aspirational: every AST node is a plain object with a string `kind` and an `at`, and `unspanned()` walks a tree and names by path every node that broke it, so a node added later without a span fails the suite that added it. The lexer is where spans come from — pull-based rather than a list-returning pass, because a passage body is raw prose between braces (B29) and needs different rules, and because a lexer that hands back one token at a time can report a bad character, step over it and carry on. The wording of each message and the warning list stay with B50, which depends on this.

**B02 · Runtime budgets as host configuration** — `limits.ts`, `budget.ts`.

Static caps refuse at compile; runtime budgets fault while running. `limits.ts` is the table of what exists, what each bounds and what exceeding it means, with the spec's two figures tables as the defaults a host starts from. The numbers are the host's: `limitsFrom` takes any of them and keeps the spec's for the rest, and refuses a figure that is not a whole number above zero — loudly, at the host's boot. Nothing is hard-coded anywhere else, and where the spec's default column is prose rather than a figure ("as the host says", "a backstop that should never fire") the default is `null` rather than a quota the language invented. There is deliberately no cap on statements in a body, and a spec holds that line. `budget.ts` is the per-turn meter B15, B23, B27, B34 and B39 will all charge against, so none of them invents a second one: it holds no figure of its own, and it never reads the clock — the wall-clock backstop checks a `now` the host passes in, at a sampling stride, so a turn stays deterministic and a suite can prove the backstop fires without waiting for it.

**B03 · Closed-bundle compile** — `sha256.ts`, `bundle.ts`, `compile.ts`.

Libraries are statically linked, so the bundle is closed and whole-bundle checking is exact rather than a guess. `sha256.ts` is written out rather than reached for: the language imports zod and nothing else and runs in a browser as readily as on a server, where `node:crypto` is not and `crypto.subtle` is asynchronous. A blessed library is a hash the host knows, so the hash is a trust boundary and a cheaper digest would let a modified library wear a blessed one's name. A library's hash is its source and nothing else, so two copies vendored under different names hash alike and a fork does not. A bundle's hash is the manifest, the world's own files and each library by name, level and hash — two bundles hash alike exactly when they would run alike, which is what the log needs beside a publish. `compile.ts` holds the two tiers: `checkShape` reads one file alone (what an editor runs on each keystroke), `compileBundle` reads the closed bundle.

**The manifest** (settled by Eric mid-PR) holds the world's name, version, author and licence; every library's version and the content hash of its source; and an enumeration of the world's own `.sprout` and `.prose` files. Two more are already there: the extensions it pins, which is the spec's own (line 1345), and the language level, which is our decision and recorded as one. This is what makes the closed bundle checkable before a line has parsed, and it turns two inferences into checks:

- **What travelled is exactly what was meant to.** A file the manifest names that did not arrive is *missing* — a reason B04 could not previously produce, because "missing" had nothing to be missing from. A file that arrived and the manifest does not name is refused at publish and flagged at load.
- **Every library is the library it says it is.** A vendored library's source is hashed against the recorded `sha`. A mismatch is a new absence reason, `mismatched`, and at load such a library is not used at all — it does not enter the bundle and its bytes are not counted. Running a world against a library it did not mean to vendor would be worse than running it without one.

Blessed library source is exempt from the source and file caps, so using the standard library costs an author nothing and a fork costs them everything.

**B04 · Strict publish, lenient load** — `absent.ts`, and the mode on `compileBundle`.

`absent.ts` carries the spec's absent table as data, because nine parts of the language consult it and each should read the same row. Under lenient load a library that did not travel or does not match its recorded hash, a file the manifest names that did not arrive, a file the host is withholding, and a file that does not compile each become a gap recorded in the bundle — with a broken file's own problems kept as warnings so the reason is legible. Two things stay refusals in both modes: text newer than the compiler, and what was never allowable at all. A static cap goes the other way, a warning at load, since refusing would darken a world accepted at publish. `softenPolicy` is the level mechanism: a diagnostic may carry the level its rule was introduced at, and at load a refusal introduced after the level the bundle was *accepted* at applies as a warning.

## What is deliberately not here

Nothing parses yet, so a compiled bundle carries no definitions and an empty word set. B05–B19 fill the definitions, B27 the word set. Where a check needs syntax that does not exist, the code names the item that brings it rather than pretending. Every such seam is a comment at the place it will be filled.

## Review

Two passes by the `pr-review` agent, both reproducing the gate green in their own worktree. `gh` is not installed here, so it returned its findings and they are posted on its behalf, each as its own comment.

**At `7f162f1`** — 2 blocking, 1 non-blocking. All fixed in `36b2a5f`:

- **Blocking** — the absent table's arrival row named `displaced` as the passage that tells a visitor the world cannot admit them. The spec names no passage there, and `displaced`'s own text is false for somebody who never stood anywhere. The row now names none; the silence is recorded; a new spec pins every passage the table *does* name to one the standard library actually declares.
- **Blocking** — `softenPolicy` was given `manifest.level` where the spec ties leniency to the level a world was *accepted* at, which is the bundle's. Invisible to the gate, and it will stay invisible until the first refusal carries a `since` — caught by reading, not by 600 passing tests.
- **Non-blocking** — extension pin diagnostics pointed at the `extensions` key rather than at the pin. Fixed rather than declined.

**At `36b2a5f`** — 0 blocking, 2 non-blocking. Both taken in `6e30cc2`, along with a third thing the reviewer raised as an aside and did not class as a finding: my working-notes entry cited the manifest's language level as the spec's own words, and it is not — that phrase is `bundle.ts`'s own comment. Only the extensions claim is spec-backed. The note contradicted the open hole two entries above it, and now does not. The reviewer also observed that using `arrived` for the byte and file counts is a *fix* rather than a refactor: at `7f162f1` a withheld file's bytes could have counted against the caps while being reported as a gap everywhere else.

## Receipts

Both at `6e30cc2`, the PR head.

```
npm run gate
```
(exit 0 — 596 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 6e30cc2)
```

## What was fenced

Nothing. No suite, corpus entry or example describes a language that stopped existing, because none of these four items replaces a behaviour the old implementation has — they add the foundation the replacement is built on. The previous language's caps in `definitions.ts` and its `microworld.ts` strict/lenient machinery both stay, and are deleted by the issues that replace the compiler around them.

Two things moved rather than being duplicated:

- `LANGUAGE_LEVEL` moves from `definitions.ts` to `bundle.ts`, re-exported from where it was, so there is one of it.
- Two names the legacy compiler still holds at the package boundary went to the new code under other names: `BundleOptions`/`BundleResult` (the legacy `CompileOptions`/`CompileResult` are internal to `sprout-lang.ts`) and `Gap` for the per-occurrence absence record (`microworld.ts` exports `Absent`, and `core/runtime.ts` reads it). The absent table keeps the spec's vocabulary throughout. The issue that deletes the legacy compiler takes both names back.

One boundary widened, visibly, and approved by Eric: `node:crypto` joins the short list of what a **spec** may import, so `sha256.spec.ts` can check the language's own SHA-256 against the platform's for a thousand random inputs as well as against the published vectors. No `src/` file imports it, and `boundary.spec.ts` still holds that line.

## Spec gaps recorded

All in the working notes' *Holes in the spec*, each decided the narrow way — the way that can be widened later.

**B01**, all lexical: what a comment looks like (`//`, inferred from the spec's own examples); what `:` means with no lowercase letter after it; what a backslash means inside quoted text; the gutter of a printed diagnostic (the spec's three examples are hand-set and disagree by a column).

**B02**: what an unset host limit means — five static caps and the wall-clock backstop have no figure, and nothing says whether a host that sets none leaves a world unbounded or refuses it; and where the three world-scoped budgets are counted, since they sit in a table whose preamble says the budgets are per turn and they are not.

**B03**: whether the file cap counts library files, since the exemption names only "the source and kind caps"; where a part's language level comes from, since a level raised by syntax cannot be derived before there is a parser; and two from the manifest — a version is free text rather than semver, and an author is a name rather than a structured record.

**B04**: which passage tells a visitor a world cannot admit them (the row says the world "says so" and names nothing); what a static cap exceeded at load means; and whether a withheld file changes the bundle's hash, given the log records a publish with the bundle's hash and a withholding as its own event.

**Closed by Eric during the PR**: what a manifest holds. The bundle also carries three things the spec's list does not name, with the reasons in the notes: the manifest it was compiled from, the size its source came to, and its own hash.

## A spec change to make

**The spec's *Files* section should say what a manifest holds.** Per CLAUDE.md I have not edited `sprout-design-spec.md`; this is the proposal. The contents are Eric's, above. Note that the spec currently mentions a manifest only twice — that a microworld is one (line 196) and that a world names its extensions in one (line 1345) — so everything else about it, the language level included, is currently a decision rather than a reading.

## One process note

`gh` is not installed in this environment, so the issue claims, this PR and the review comments went through the GitHub API rather than `gh pr create`. The gate and e2e ran locally, as the rules require. The four issues landed on one branch rather than four, at Eric's direction when asked at the start of the session.


ericeslinger:
**Blocking** — `sprout/lang/src/absent.ts:74-77`

The `place-of-arrival` row sets `told: 'displaced'`, asserting the world's `displaced` passage is what tells a visitor the world could not admit them — a resolution the spec does not state and that isn't recorded as a hole.

The spec's row for this case ("a place the world says visitors arrive at | the world does not admit anyone, and says so") never names a passage, unlike the adjacent `place-underfoot` row, which explicitly says "told through the world's `displaced` passage." A visitor who was never admitted was never "standing" anywhere, so `displaced`'s own text — "The place you were standing is gone." — reads oddly for them; and "The host guarantees [a nickname] exists before the world ever sees them, so the language never has to render an absence" suggests this may be a host-level refusal with no passage involved at all, not a Sprout passage. `absent.spec.ts:42-43` locks the guess in as fact ("tells a displaced visitor, and a visitor a world cannot admit"). Either record this as a hole in the working notes' B04 section (which already records two others) or leave `told` unset until it's answered.

*Posted on behalf of the `pr-review` agent, which reviewed at `7f162f1`. `gh` is not installed in this environment, so the agent returned its findings rather than posting them itself.*

---

ericeslinger:
**Blocking** — `sprout/lang/src/compile.ts:401`

`compileBundle` softens load-mode refusals against `manifest.level` — the world's own declared level — not the bundle's actual level, and the two can differ; the spec ties "accepted at one level" to the bundle's level, not the manifest's.

The Language levels section says in one breath: "a bundle's level is the highest of any of its parts, library source included. A world accepted at one level keeps loading when the language tightens." `compileBundle` computes exactly that maximum as `level` (line 317-320) and stores it as `bundle.level`, but the `softenPolicy` call at line 401 passes `manifest.level` instead. This is reachable today, not hypothetical: compile.spec.ts's "takes a library's level when the library is newer" test accepts a bundle with `manifest.level: 1` and `bundle.level: 2`. Once any refusal carries a `since` (the whole point of `since` and `softenPolicy`, per B04), a world using a level-2 library would have a `since: 2` refusal wrongly softened to a warning at load, even though the bundle itself already runs at level 2. It's dormant only because nothing yet sets `since` — compile.spec.ts even has a test titled "has nothing to soften yet" — so nothing catches it. Not recorded as a hole, despite the working notes' adjacent "where a part's language level comes from" entry covering the same area. Pass `level`, not `manifest.level` — or, if `manifest.level` is deliberate, record why.

*Posted on behalf of the `pr-review` agent, which reviewed at `7f162f1`.*

---

ericeslinger:
**Non-blocking** — `sprout/lang/src/compile.ts:220-233`

Every diagnostic about an extension pin — a bad major version, or a repeated name — points at the same span: the first occurrence of the literal `"extensions"` in the manifest text, from `atKey(manifestFile, 'extensions')`, not the specific pin's own location.

A manifest with more than one pin would report a problem with the second or third pin at the first pin's position. Doesn't show with the worked example's single extension, and is reasonable scaffolding before the manifest has real per-field spans, but falls short of "the token" the spec's Diagnostics section asks for.

*Posted on behalf of the `pr-review` agent, which reviewed at `7f162f1`.*

---

ericeslinger:
**Review**: 2 blocking, 1 non-blocking. Gate at 7f162f1: green (exit 0, last line `✓ corpus: every good archive passes, every bad one fails as expected`). Solid B01–B04 foundation — sha256.ts checked correct at the 2^29-byte boundary against node:crypto (analytically and by actually hashing ~512MiB), limits.ts matches the spec's two tables exactly, and the boundary widening for node:crypto is properly spec-only — but the load-mode "which level was this accepted at" threshold uses the wrong value and one absent-table row asserts an unrecorded guess about which passage runs.

*Posted on behalf of the `pr-review` agent, which reviewed at `7f162f1`.*

---

ericeslinger:
## Answering the review, and one change from Eric

All three findings are fixed in `36b2a5f`, which also carries a decision from Eric on the manifest that landed while the review was running.

### Blocking — the absent table's arrival row named `displaced`

**Fixed, and the reviewer is right about more than it claimed.** I checked the three `told` values against the spec rather than against my memory of it:

- `place-underfoot` → `displaced` — spec line 1456 names it outright, and line 1643 declares `passage displaced default { The place you were standing is gone. }`. Correct.
- `extension` → `missing` — spec line 1345 says the visitor is told "through the world's `missing` passage", and line 1642 declares it. Correct. (The review didn't flag this one, also correctly.)
- `place-of-arrival` → the spec says the world "does not admit anyone, and says so" and names nothing. My `displaced` was a guess, and the review's argument for why it's the *wrong* guess is the decisive part: nobody who was never admitted was standing anywhere, so that passage's own text is false for them.

The row now carries `told: null`, with `null` re-documented to mean "the spec names no passage" rather than "nobody is told" — the row still says the world says so. The silence is recorded in the working notes' B04 group. I also added a spec that pins every passage the table *does* name to one the standard library actually declares, so the next invented passage name fails a test instead of shipping.

### Blocking — softening against `manifest.level` rather than the bundle's

**Fixed.** `softenPolicy` now takes `level`. The reviewer's reading of the spec is the right one: "a bundle's level is the highest of any of its parts" and "a world accepted at one level keeps loading" are the same sentence, so the acceptance level is the bundle's, not the one the manifest was written for.

Worth saying plainly: this was invisible to the gate and will stay invisible until the first refusal carries a `since`, which is exactly why a reader caught it and 601 passing tests did not. I've added a spec pinning the number the threshold is taken from (`bundle.level` is 2 for a world whose manifest says 1 and which vendors a level-2 library), which is the most that can be asserted before a real policy refusal exists — the wiring itself still gets its first end-to-end exercise from the first `since`.

### Non-blocking — extension diagnostics pointing at the key

**Fixed** rather than declined; it was two lines. Both the bad-major-version refusal and the repeated-pin refusal now point at the pin's own name via `atValue`, the same helper the library pins and the file list already used. The remaining limitation is real and now documented on `atValue`: it finds the *first* place a value is written, which is right for a name that appears once and is scaffolding until the manifest is read with spans of its own rather than arriving already parsed.

### And the manifest, from Eric

Eric settled the largest of the recorded holes — *what a manifest holds* — while the review was in flight: **name, version, author, licence, library versions and shas, and an enumeration of the local `.sprout` and `.prose` files.** Two more are the spec's own and stay: the language level (*"a microworld's manifest records the level it was written for"*) and the extensions it pins (spec line 1345: *"a world names the extensions it uses, pinned by major version, in its manifest"* — which I'd flagged as an addition beyond his list and turns out to be spec-backed).

This is not bookkeeping. It turns two things the compiler was inferring into things it checks, neither of which was catchable before:

- **What travelled is exactly what was meant to.** A file the manifest names that did not arrive is *missing* — a reason B04 previously had no way to produce, because "missing" had nothing to be missing *from*. A file that arrived and the manifest does not name is refused at publish.
- **Every library is the library it says it is.** A vendored library's source is hashed and checked against the recorded `sha`. A mismatch is a new absence reason, `mismatched`, and at load such a library is **not used at all** — it doesn't enter the bundle and its bytes don't count. Running a world against a library it did not mean to vendor would be worse than running it without one.

The bundle now carries its manifest rather than a lone world name: one field instead of four, and it's what the bundle was compiled from anyway. Two smaller silences remain and are recorded — a version is free text rather than semver, and an author is a name rather than a structured record.

**The spec's *Files* section should gain this.** Per CLAUDE.md I haven't edited the spec; the change is proposed here for Eric to make.

Eric also approved the `node:crypto` boundary widening, so `sha256.spec.ts` keeps its differential check against the platform. No `src/` file imports it and `boundary.spec.ts` still holds that line.

### Receipts, re-run at the new head

Both at `36b2a5f`. The previous receipts named `7f162f1`, which is no longer the head, so they are replaced rather than carried over.

```
npm run gate
```
(exit 0 — 594 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 36b2a5f)
```

Still nothing fenced. The `pr-review` agent is being re-run at `36b2a5f`, since the manifest work is a substantial change it has not seen.

---

ericeslinger:
**Non-blocking** — `sprout/lang/src/compile.spec.ts:440`

The load-mode test for a broken *library* file (as opposed to a broken world file) was removed in this diff and not replaced, so the "loading is lenient" describe block no longer directly exercises `reason: 'broken'` for a library's own file.

The underlying code is unaffected — the tier-one shape-check loop (compile.ts:508) walks `[...arrived, ...usable.flatMap(l => l.files)]` with no branch distinguishing world files from library files, so the existing world-file broken test already exercises the same path a library-file version would. Worth restoring for directness, given the publish-mode sibling test right above it (compile.spec.ts:370-378) already shows the small manifest change needed (a pin with `sha: libraryHash(broken)` so the library is `usable` and reaches the syntax check).

*Posted on behalf of the `pr-review` agent, which re-reviewed at `36b2a5f`.*

---

ericeslinger:
**Non-blocking** — `sprout/lang/src/bundle.ts:80`

`Manifest.files`'s doc comment says "What travelled must be exactly this," but that's only enforced at publish; at load, a file that arrives without being named in the manifest is warned about and still runs (compile.ts:354-361, `report.strict`), not excluded.

A reader of the type alone would expect the invariant to hold always. Tightening the comment to say the equality is enforced at publish and merely flagged at load would match what compile.spec.ts:432-437 ("warns rather than refuses about a file the manifest does not name") actually tests.

*Posted on behalf of the `pr-review` agent, which re-reviewed at `36b2a5f`.*

---

ericeslinger:
**Review**: 0 new blocking findings, 2 non-blocking. Gate at 36b2a5f: green (exit 0, last line `✓ corpus: every good archive passes, every bad one fails as expected`). All three prior blocking findings fixed properly (recorded holes and corrected values, not patched symptoms); the manifest/mismatched-library rework is internally consistent — `usable` is threaded correctly everywhere it needs to be, the withheld/missing gap interaction produces exactly one gap, and `canonicalManifest`'s use of `JSON.stringify` sidesteps the seam-crossing risk the old `hashOfNamed`-based scheme had to solve by hand.

*Posted on behalf of the `pr-review` agent, which re-reviewed at `36b2a5f`. It also confirmed the spec file is untouched (`git diff main...HEAD -- docs/design/sprout-design-spec.md` empty), and that every file it cleared in round one is byte-identical between `7f162f1` and `36b2a5f`.*

---

ericeslinger:
## Answering the re-review

Both non-blocking findings taken, plus a third thing the reviewer raised as an aside that turned out to be a real error in what I wrote. All in `6e30cc2`.

### The aside, which was the important one

The reviewer noticed, and did not even class as a finding, that my working-notes entry cited *"a microworld's manifest records the level it was written for"* as the spec's own words. It is not. The design spec mentions a manifest exactly twice — line 196 (a microworld *is* one) and line 1345 (a world names its extensions in one). The phrase I quoted is `bundle.ts`'s own comment, by way of the README.

So of the two fields I said were "already there by the spec", only the extensions actually are. The language level is in the manifest by *our* decision, which is precisely what the open hole two entries above it already said — the two notes contradicted each other and I did not notice. Fixed: the note now separates the one the spec gives from the one we decided, and says the latter is a decision and not a reading.

Worth stating plainly because the whole point of the *Holes in the spec* discipline is that it is honest about what the spec does and does not say, and an entry that launders a decision into a citation is worse than no entry.

### Non-blocking — `Manifest.files` promised more than load enforces

**Taken.** The comment said "What travelled must be exactly this," which is true at publish and false at load, where an unnamed file is flagged and still runs. It now says which half is enforced where, and keeps the asymmetry explicit: a file named here that did *not* travel is absent either way.

### Non-blocking — the load-mode broken-library-file test

**Taken.** The reviewer is right that the code is unaffected — the tier-one pass walks `[...arrived, ...usable.flatMap(l => l.files)]` with no branch between world files and library files, so the world-file case already covers it. But that is only obvious once you have read the loop, and the test is three lines. Restored as its own describe with both halves, publish and load, and a comment saying why it exists when its sibling covers the same path.

### On the reviewer's own correction

It flagged that `arrived` being used for the byte and file counts is a **fix** rather than a refactor — at `7f162f1` those counted from `source.files` while the hash already used `arrived`, so a withheld file's bytes could have counted against the caps while being reported as a gap everywhere else. That inconsistency was mine and it went in unnoticed; it is gone now as a side effect of the manifest work rather than by design. Recording it here because "fixed by accident" is worth knowing about.

### Receipts

Both at `6e30cc2`, the PR head. The `36b2a5f` receipts are replaced.

```
npm run gate
```
(exit 0 — 596 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 6e30cc2)
```

No blocking findings outstanding. Still nothing fenced.

---

PR #57, opened 2026-09-21
spec: what a manifest holds

ericeslinger:
The spec names a manifest as one of the four things a microworld is made of and never says what is in one. #56 needed rather more of it than that, so Phase 0 shipped a manifest and recorded its contents in the working notes as a decision standing outside the spec. Eric settled the contents and asked for the spec change; this is it.

No code changed. `git diff` is two documents.

## Where, and why there

**`The world model › The manifest`**, a new subsection under *Files*.

*Files* is the only place that claims to say what a microworld is made of, and the sentence naming the manifest is the incomplete one — so the gap is literally there, in the section a reader hits first when asking "what do I have to write?". It is also early and author-facing, where *The compiler* is late and ordered for someone checking the design.

The manifest's fields cut across three other sections — the level belongs to *Language levels*, the extension pins are already stated under *Extensions*, the library hashes to *Kinds › Libraries and namespaces*. Rather than restate any of them, the new subsection names all eight fields and points out to the sections that own the detail. That is the pattern *Files* already uses for the absent table ("What 'absent' means at each kind of reference is a table under The compiler").

The alternative considered was a subsection under *The compiler*, paired with *What compiling produces* — input beside output. Rejected because *Files* would still have to say something, leaving two half-statements of the same thing.

## What it now says

The eight fields as a table, then the two things the manifest makes checkable that nothing else could:

- **`files`** is what makes a file that did not arrive *missing* rather than merely absent — without it there is nothing for a file to be missing from. What travelled must be exactly what the manifest names; publishing refuses otherwise, and loading says so and runs.
- **`libraries`** is what closes the bundle. A library named there whose source did not travel is absent, and so is one whose source does not hash to the recorded value: a world runs against the library it meant to vendor or against none, because running against some other copy under a trusted name is the worse of the two failures.

And the line that earns the section: *none of it needs a parser, and that is the point — whether a bundle is closed and whether it is complete are both settled before a line of the language has been read.*

## Two clauses elsewhere

Both where the manifest was already implied and merely unnamed:

- **`What compiling produces`** lists the manifest among what a bundle carries, which it did not.
- **`Language levels`** says where each part's level is declared — the manifest for the world, its own source for a library — and that the level a world was *accepted* at is the bundle's, not the one its manifest asked for. That second half is the distinction the review of #56 caught the compiler getting wrong; one sentence here means the next reader does not have to re-derive it from two paragraphs that were previously a page apart.

## What I did not write in

Three things in the shipped code that are mine rather than Eric's, left out so that the spec does not quietly acquire them:

- that a version is free text rather than semver;
- that an author is a name rather than a structured record;
- the `mismatched` absence reason as a distinct name (the spec says the consequence, not the vocabulary).

All three stay in the working notes as narrow decisions. Say the word and any of them can go in.

## Working notes

The entry recording this as proposed-not-made is replaced by one saying the spec now carries it. The *where a part's language level comes from* hole narrows rather than closing: where a part **declares** its level is now stated, and what remains open is that nothing yet checks a declared level against the syntax actually used — "added syntax raises it" describes a derivation no one can perform until there is a parser that knows which syntax raised it, so today a part's level is taken on trust.

## Not a backlog issue

Per CLAUDE.md, work that is not a backlog issue says in its PR why it is too small to be one. This is the documentation half of #3, which is closed: the decision was made during that issue's PR, and the spec edit was deferred only because CLAUDE.md forbids the coding agent making it unasked. Opening an issue to hold a paragraph Eric has already written the content of would be ceremony.

## Receipts

Both at `bda89be`, the PR head. Documents only, but the rule is unconditional.

```
npm run gate
```
(exit 0 — 596 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at bda89be)
```

Nothing fenced.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UDcfUL7NKgoiU8Spjas1Su)_


ericeslinger:
**Blocking** — `docs/design/sprout-working-notes.md:212`

The paragraph claiming the spec's bundle-field list doesn't name the manifest is now false — this same PR adds exactly that to the spec two hunks below.

At `bda89be`, working-notes.md:212 still reads "The bundle carries three things the spec's list does not name: the manifest it was compiled from, ...; the size ...; and the bundle's own hash...". But sprout-design-spec.md:1521 ("What compiling produces"), edited by this PR, now opens "A bundle: the manifest it was compiled from, the definitions, ...". A reader following the working notes' claim to the spec finds it contradicted by the PR that made the claim stale.

Fix: change "three things" to "two things" and drop the manifest clause from the enumeration — size and the bundle's own hash are the two the spec's list still doesn't name.

*Posted on behalf of the `pr-review` agent, which reviewed at `bda89be`. `gh` is not installed in this environment, so the agent returned its findings rather than posting them itself.*

---

ericeslinger:
**Non-blocking** — `docs/design/sprout-design-spec.md:217`

The cross-reference "under Kinds › Libraries and namespaces" uses a breadcrumb style that appears nowhere else in the document.

Every other internal pointer in the spec names a single section bare: "in Names" (line 196), "under The compiler" (line 219), "under The runtime" (line 1624). Before this PR, "›" appears zero times in either sprout-design-spec.md or sprout-working-notes.md; this PR introduces it here and once more in the working notes. Reads as carried over from this PR's own prose register rather than the document's.

Fix: "under Libraries and namespaces" matches the established convention.

*Posted on behalf of the `pr-review` agent, which reviewed at `bda89be`.*

---

ericeslinger:
**Non-blocking** — `docs/design/sprout-design-spec.md:219`

The paragraph "A file that is removed, withheld, named in the manifest and not delivered, or that fails to compile reads as absent..." now sits nested under the new "### The manifest" heading, though its content is Files' general closing statement, not specific to the manifest.

No heading separates it from "### The manifest" (line 198) before "## Kinds, composition and libraries" begins (line 221), so by the document's own heading structure it now reads as part of the manifest subsection. But "removed", "withheld" and "fails to compile" apply independently of the manifest (withheld files aren't even manifest-tracked — they're `source.withheld`), and the closing pointer to the absent table is Files' own capstone, carried over unchanged from before this PR existed.

Fix: move it above "### The manifest", right after the opening "A microworld is a manifest..." sentence, where it sat before this PR.

*Posted on behalf of the `pr-review` agent, which reviewed at `bda89be`.*

---

ericeslinger:
**Non-blocking** — `docs/design/sprout-working-notes.md:214`

The sentence added to the "Closed on" paragraph now says the same thing twice, four lines apart.

"Two smaller things stay decided rather than stated: a version is free text rather than semver, and an author is a name rather than a structured record" (line 214, new) duplicates the unedited "Still unsaid, and decided the narrow way: whether a version is semver or free text (free text, non-empty), and whether an author is a name or a structured record (a name)" (line 218).

Fix: keep one; the "Still unsaid" paragraph already carried this before the PR.

*Posted on behalf of the `pr-review` agent, which reviewed at `bda89be`.*

---

ericeslinger:
**Review**: 1 blocking, 3 non-blocking. Gate at bda89be: green (exit 0; 596 passed | 7 skipped sprout, 40 passed cli; last line `✓ corpus: every good archive passes, every bad one fails as expected` — matches the PR's receipts). Doc-only change: the new manifest table and both trailing clauses check out accurately against `Manifest`/`Bundle`/`compileBundle` at main (9b551ba); the one blocking item is a working-notes paragraph the PR's own spec edit makes stale two hunks away.

*Posted on behalf of the `pr-review` agent, which reviewed at `bda89be`. Its brief for this PR had the "a spec diff is blocking" rule suspended (Eric asked for the change) and replaced with a field-by-field check of the new text against the code at `main`, plus a check that nothing beyond what Eric settled had been written in.*

---

ericeslinger:
## Answering the review

All four taken, in `d71f3fc`. I verified each against the files rather than on the reviewer's word, and all four hold.

### Blocking — the notes' claim that the spec's list omits the manifest

**Fixed, and it is the best kind of finding.** Adding "the manifest it was compiled from" to the spec's bundle list made my own working-notes sentence — "the bundle carries three things the spec's list does not name: the manifest…" — false, in the same diff, two hunks away. Neither half is wrong on its own; the contradiction only exists because both moved at once, which is exactly what a reader catches and a writer does not.

It now reads *two* things: the size its source came to, and the bundle's own hash. (The spec's list names "the hash of every vendored library", which is a different hash from the bundle's own, so that one stays.)

I first fixed it with a parenthetical explaining that the manifest is now named in the spec, and then deleted the parenthetical: the spec is authoritative for the manifest as of this PR, so the notes have no business relitigating it.

### Non-blocking — the `›` cross-reference

**Taken.** `Kinds › Libraries and namespaces` uses a breadcrumb that appears **zero** times in either document before this PR — I checked `git show origin/main:docs/design/sprout-design-spec.md | grep -c '›'`. The spec points at a section bare: "in Names", "under The compiler", "under The runtime".

The breadcrumb style is real, but it belongs to the *backlog* and the issues, which name spec sections from outside. Inside the spec there is no path to walk. Both occurrences are gone — the one in the spec and the one I had put in the working notes.

### Non-blocking — Files' capstone nested under the wrong heading

**Taken, and it is a genuine structural error rather than a nit.** Inserting `### The manifest` before Files' closing paragraph silently reparented it: "a file that is removed, withheld … reads as absent" and the pointer to the absent table became, by Markdown nesting alone, part of the manifest subsection. The reviewer's argument for why that is wrong is the right one — `removed`, `withheld` and `fails to compile` have nothing to do with the manifest, and a withheld file is not even manifest-tracked (it is `source.withheld`).

The paragraph is back above the subsection, where it was. It keeps the clause I added to it ("named in the manifest and not delivered"), which is a forward reference to the subsection below — acceptable, since the sentence two lines above it already names the manifest.

### Non-blocking — the duplicated sentence

**Taken.** My new sentence about versions being free text and authors being names sat four lines above the paragraph that already said it, with the older one giving more detail ("free text, non-empty"). Mine is gone.

### What the reviewer cleared, which is the part that mattered

Its brief was to check the new text field-by-field against `Manifest`, `Bundle` and `compileBundle` at `main`, and to flag anything asserted in the spec that was only my implementation choice. It reports all eight fields matching the interface in the same order; the `files` paragraph true in both directions at both `report.gap` and `report.strict`; the `libraries` paragraph holding for both `missing` and `mismatched`; the *Language levels* clause stating the accepted-level distinction correctly rather than approximately; and the three things I said I had kept out — semver, structured authors, and the `mismatched` vocabulary — genuinely absent from the spec.

That last check is the one I could not do for myself, since the whole risk of an agent editing a spec is that it writes down what it built and calls it design.

### Receipts

Both re-run at `d71f3fc`, the PR head. The `bda89be` receipts are replaced.

```
npm run gate
```
(exit 0 — 596 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at d71f3fc)
```

No blocking findings outstanding. Nothing fenced. Still no code changed.

---

PR #60, opened 2026-09-21
#59: the recovery paths in parse.ts, restructured

### LINKED ISSUE #59: A structural pass on `parse.ts` recovery, before Phase 2 builds on it
Not a backlog item. This comes out of #58, where the `pr-review` agent made the same call twice and Eric took it.

## Why

`parse.ts` recovers from a declaration it cannot read so that one bad declaration costs that declaration and not the file. The recovery is hand-rolled lookahead with several interacting boolean checks, and over #58 it produced five bugs in three rounds:

| round | what broke |
| --- | --- |
| 1 | an unclosed enum read the `enum` that started the **next** declaration as an option, ate the token recovery needed to resynchronise, and swallowed that declaration whole — with no diagnostic naming it |
| 1 | a failed type parse read as *no type written*, so the keyword `default` was swallowed as a value and a property already reported as broken was returned anyway |
| 1 | a character the lexer steps over leaves no token, so the parser blamed a missing comma for a gap it could not see |
| 2 | the fix for the first made a **well-formed** enum using `message` as an option unreadable — and contradicted a note added in the same commit saying option names are not reserved |
| 2 | the fix for the third checked "did the lexer skip something here?" *before* trying to consume the comma, so a comma the author had written was blamed and what followed it was lost |

Every one is fixed and specced at `adeeadd`. The pattern is the finding:

> hand-rolled lookahead/recovery with several interacting boolean checks is exactly where "fixed the reported input, broke an adjacent one" keeps happening here, twice in a row
>
> — `pr-review`, on #58

Two of five were introduced *by* a fix written in direct response to a hostile review of that same logic. That is not a competence problem with any single fix; it is a shape problem with the code.

## Why now

#12 (B12, the world root) and #19 (B19, composition) add the kind and object parsers, which is where the recovery logic grows most — nested bodies, several member kinds, more places a brace can be forgotten. **The cheapest moment for a structural look is before those land**, not after they have doubled the surface.

## What this is

A `/code-review` pass over `parse.ts` and `lexer.ts` at high effort, then act on it. The likely shapes of an answer, in no particular order:

- Replace the ad-hoc "is this a separator / a keyword / a gap" checks with one explicit decision about what token the parser is looking at and why.
- Make recovery a named strategy per construct rather than two methods (`recover`, `recoverInBraces`) with subtly different rules about what stops them.
- Consider whether the lexer should hand the parser a token for a character it refused, instead of stepping over it and leaving `refusedBetween` to reconstruct the gap after the fact. That would make the parser's view of the input complete, which is what the third and fifth bugs were both really about.

Not in scope: new syntax, or changing what the parser accepts. This is about the recovery paths, and any change should leave `parse.spec.ts`'s existing assertions passing unchanged.

## How to know it worked

The paired-table specs added in #58 (`enum`/`message` as an option vs. a forgotten brace; a stray character beside a written separator vs. without one) stay green, and adding a construct in #12 or #19 does not need a new special case in the recovery logic to avoid eating the declaration after it.

ericeslinger:
Closes #59

A `/code-review` pass over `parse.ts` and `lexer.ts` at high effort, as the issue asked, then acting on it. **Fourteen findings.** Three are the shape of the code; eleven are what that shape produced. Three rounds of `pr-review` on top of that found four more, all fixed.

No new syntax, nothing the parser accepts has changed, and every assertion that existed before this PR still passes — with one deliberate exception, noted at the bottom.

## The shape

**One table instead of four answers.** Four places asked *"does a declaration start here?"* and gave three different answers: `file()` dispatched on two string literals, `recover()` matched a list by spelling, `atDeclarationKeyword()` matched the list plus a lookahead. That disagreement *is* the original recovery bug from #58, still in the file. They now read one table of word → reader — and the "this compiler reads …" message is built from that same table, so the two cannot drift.

**A token that says for itself whether a character was refused before it.** `refusedBetween` reconstructed, after the fact and from offsets, a gap the lexer had made by stepping over a character. It had been wrong three times in three directions — too narrow (#58's bug 3), consulted at the wrong point (#58's bug 5), and too *wide*, because a caller's node span was six characters short. That last one is worth stating plainly: **a defect in `propertyBody`'s span turned into a wrong diagnostic in `remembers()`**, which is what a mechanism that depends on callers passing exact offsets buys you. The reconstruction is gone. `Token.afterRefusal` carries it instead: no offsets to pass, none to go stale, and the linear scan that ran once per option goes with it.

I took the flag rather than the `bad` token kind the issue floated. A `bad` token would let the parser do something *different* at a bad character, and there is nothing different to do — it is always skipped, and the only question is whether it explains a gap. The review tried to find an input where the parser needs to know *where* the character was and could not.

**A missing separator is reported only once the next item reads.** A token that cannot be an item at all is that token's problem. Charging it as a missing comma as well told the author to add a comma that would not have helped.

## What that fixed, and what else did

**Eight diagnostics that were not the author's mistake:**

| input | said | now says |
| --- | --- | --- |
| `enum { message, silver }` | *An enum needs a name* **and** *A message needs a name* | the first only |
| `enum Ward {` | *is never closed* **and** *has no options* — and the reading order sorts by offset, so the author read the wrong one **first** | the first only |
| `enum Ward { oak 4 }` | *needs a comma* **and** *cannot hold the number 4* | the second only |
| `enum Ward { oak, message` | *is never closed* **and** *A message needs a name* | the first only |
| `enum Ward { 4` ⏎ `message :x` ⏎ `}` | *is never closed* at the end of the file, past the brace | at the keyword where recovery stopped |
| `message :n2 with boolean` ⏎ `. message :n3 …` | *`boolean.` is not followed by a name* — and `:n2` silently gone | the stray `.` named, both declarations kept |
| `:x [Ward, oak]` | *A list type is never closed* — about a bracket written correctly | *A list type names one element type* |
| `:remembers [a: 1.5, b: 2]` | *needs a comma* — the author wrote the comma | *Sprout has no fractions* |

**Four declarations or items that went missing:** content nested inside an unmatched brace promoted to the top of the file (`enum Ward { 4 { message :x } }` produced a top-level `message :x` the author wrote two braces deep); the `boolean.` case above; and a well-formed list element or `:remembers` entry following one that could not be read (`:x [oak, Zeta silver]` kept only `oak`). The list and `:remembers` loops read *on* past a bad item now rather than skipping to the next comma, which is what their own comment already promised.

**One crash.** `typeExpr` and `literal` recursed once per bracket with no guard, so ~7,000 of them threw `RangeError` out of a parser whose own docblock says *"nothing is thrown"* and whose spec asserts it three times. They count against the host's `nesting` cap now — which means the first tier had to be *given* the caps rather than assuming them, since a limit the language invented is the one thing Limits says the language never does.

**One double report.** A nested, unreadable, never-closed list said *"This list is never closed"* once per level of nesting around it, because the recovery `continue`d back through a loop whose own end-of-file branch said it again. It says it once.

**One inconsistency.** A list and a `:remembers` abandoned the whole construct at the first missing comma, where an enum read on. All three read on now.

**Still true, and stated rather than papered over:** the enum's own recovery drops options after a bad one, because brace-structured garbage needs a brace-aware skip and `enum Ward { 4 { message :x } }` is why `recoverInBraces` exists.

## Method

Worth recording, because it is why these were found: every pass **ran the built parser against adversarial input** rather than hand-tracing. The `/code-review` pass used 4,000 random token sequences and 32,000 mutation trials; the three `pr-review` rounds used ~13,000, ~20,000 and 8,528 generated cases plus differential oracles. Findings were reachable only that way. Hand-tracing had by then missed real bugs in this file three times.

It also taught the opposite lesson, three times in a row: **a clean run is only as good as its input space.** One oracle returned 0/16,000 because its token pool had no brackets; a second returned 0/7,574 because its nested shapes lacked an enclosing one; and the enumeration added here to cover that gap was itself vacuous for a sixth of its own table, because `atType()` routes `[[[Zeta` to the type path. So the suite now carries a **deterministic generator with brackets in the pool** (asserting classes — never throws, never says one thing twice in one place, same source same answer) and an **enumeration of the nested-bracket shape whose first test asserts that it reaches the code it is about**, by checking that no shape produces the type path's vocabulary and that the count reached equals the size of the table. That test fails against the enumeration's own first draft.

## One assertion deliberately changed

`parse.spec.ts`'s *"never reads a declaration keyword as an option"*, added in #58, is gone — replaced by two tests that state the real rule. Its expectation encoded the very rule that finding 9 identifies as wrong: it asserted that a keyword at end-of-file is a declaration, which is what produced the double refusal. Flagged because changing an assertion that pins a previous fix is exactly the move I got wrong in #58; the reviewer judged the replacement right by executing the adjacent combinations rather than reading the justification.

## One red gate, recorded

`afebfc1` was committed and pushed with `typecheck:spec` failing. A spec helper took `fn = parseProperty` as a default, which types the parameter to that signature, so passing `parseRemembers` did not typecheck — **and vitest strips types, so all 818 tests passed.** I read the test counts and not the exit code. `4e913d0` is the one-line fix, as its own commit rather than an amend, so the record shows it happened. It is in the body because a gate that runs on my machine is only worth what I admit about it.

## Receipts

Both at `11eb88b`, the PR head.

```
npm run gate
```
(exit 0 — 822 passed | 7 skipped across lang/core/stores, 40 passed in cli; 786 before this PR)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 11eb88b)
```

## Spec gap recorded

One, in the working notes: **what "expression and block nesting" counts.** The static caps table caps nesting at 8 and does not say whether a list type or a list value is a thing that nests. The parser must bound its own recursion either way, and it counts brackets against the host's cap rather than a number of its own.


ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:312`

`recoverInBraces()`'s only caller reports "never closed" at the end of the file even when recovery actually stopped at a specific, later declaration keyword, so the diagnostic points at the wrong token.

`recoverInBraces()` returns `false` for two different reasons — it ran out of file, or it found a declaration keyword at depth 1 and gave up without consuming it — but its one call site (`if (!this.recoverInBraces()) unclosed(this.source.endSpan);`) always uses `this.source.endSpan`, ignoring which one happened. For `enum Ward { 4\nmessage :x\n}\n` this reports "`Ward` is never closed." at `k.sprout:4:1` (true EOF, past the file's own trailing `}`) instead of `k.sprout:2:1` (where `message` sits, which is where recovery actually stopped), and additionally produces an unrelated knock-on "Sprout does not know what to do with \"}\" here." for the enum's own now-orphaned closer. The sibling branch three lines above (line 299, `unclosed(this.peek().at)`) gets the identical situation right when it's reached without a preceding bad option. The existing test for this call site, `'says an enum is unclosed even when a bad option sent it into recovery'` (`enum Ward { oak, 4`), only exercises the true-EOF outcome, where `endSpan` happens to be correct — it doesn't cover the keyword outcome.

Fix: `if (!this.recoverInBraces()) unclosed(this.done ? this.source.endSpan : this.peek().at);` — `recoverInBraces()` leaves the keyword token unconsumed on its keyword-triggered `false`, so `this.peek()` is exactly it.

---

**Non-blocking** — `sprout/lang/src/parse.ts:53` (and the `readers` map at 99–102, used at 241)

`DECLARATIONS` (used only to build the "reads enum and message" text in the unknown-word message) and `readers` (the actual dispatch table) are two independently-maintained lists; the header comment claims "a spec holds that the two agree," but only one direction is actually tested.

`'reads every word it says it reads'` (parse.spec.ts:748) iterates `DECLARATIONS` and would catch a word listed there without a working reader. It would not catch the reverse: a reader added to the `readers` map without a matching `DECLARATIONS` entry would parse correctly on its own, while the "Sprout does not know what to do with X here... reads enum and message" message would quietly go stale for every other unknown word. Nothing ships broken today; this is just narrower protection than the comment claims.

---

**Non-blocking** — `sprout/lang/src/parse.ts:547` (and the equivalent in `remembers()` at line 677)

When an unreadable list element (or `:remembers` entry) is not immediately followed by a comma, `skipToSeparator` swallows everything up to the next comma or `]` — including any well-formed elements in between — with no diagnostic ever mentioning what was dropped.

`parseProperty(SourceFile, ':x [oak, Zeta silver]')` returns a property whose default list has exactly one element (`oak`); `silver` is silently absent from both the AST and the diagnostics, which read only `` `Zeta`, which starts with a capital is not a value. `` `:remembers [a: 0, b c: 1]` drops `c: 1` the same way. This runs against the comment directly above `listLiteral`'s bad-element branch ("an author owed three problems is owed all three, so the list reads on rather than ending at the first bad element") — here it reads *past* silently rather than reading *on*. It isn't a new failure mode: `enumDeclaration` has always done the same thing once `recoverInBraces` lands exactly at the closing brace (a bad option mid-enum silently drops every option after it), and the new list/remembers comment explicitly frames this recovery as matching enum's ("as the enum does") — so judged non-blocking, an extension of an accepted pre-existing limitation rather than a new one. Flagging because the stated rationale oversells it for this specific case.

---

**Review**: 1 blocking, 2 non-blocking. Gate at e706bfa: green, exit 0, last line `✓ corpus: every good archive passes, every bad one fails as expected` (805 passed | 7 skipped in sprout, 40 in cli) — matches the PR's receipts; spec diff against main is empty; e2e not re-run (diff confined to `sprout/lang/src` + working notes, author's receipt stands). ~13,000 generated fuzz/mutation cases plus three differential oracles over the deferred `missingComma` (enum/list/remembers, 6,000 each) found no crash, hang, dropped/duplicate/stale comma report, or nondeterminism; the `readers` table's actual dispatch and the `afterRefusal` flag both check out — I could not construct an input where the parser needs to know *where* a refused character was rather than merely that one occurred. The one blocking finding is a diagnostic-span bug in `recoverInBraces`'s caller (only one call site, not two — its two *return* branches are what disagree), in the input adjacent to what #59 fixed there.

*Posted on behalf of the `pr-review` agent, which reviewed at `e706bfa`. `gh` is not installed in this environment, so it returned its findings rather than posting them itself. It also judged the deliberately-changed assertion right, having executed the adjacent combinations rather than taking the PR body's word for it, and said it did not think this change needs a further `/code-review` pass on top.*

---

ericeslinger:
## Answering the review

All three taken, in `1896bac`. Both reproducible findings were reproduced before being touched.

### Blocking — the unclosed enum reported at the wrong place

**Fixed as suggested.** `recoverInBraces` fails for two reasons and its caller knew only one of them, so it reported at the end of the file when recovery had stopped at a keyword two lines earlier.

```
enum Ward { 4        before                          after
message :x             1:13  cannot hold the number 4  1:13  cannot hold the number 4
}                      4:1   `Ward` is never closed.   2:1   `Ward` is never closed.
                       3:1   does not know about "}"   3:1   does not know about "}"
```

Worth naming what this is: **the third time on this file that a fix's own adjacent input was the one still broken.** The #59 pass added the `unclosed()` call at this site and covered the EOF outcome; the keyword outcome — the one the same pass had just taught `recoverInBraces` to produce — went uncovered. There is a spec for each branch now.

The knock-on `}` stays. After recovery stops at a keyword, the enum's own closing brace really is orphaned at the top of a file, and *"Sprout does not know what to do with `}` here"* is true of it. Two genuine problems and one orphan is the right count for that input.

### Non-blocking — `DECLARATIONS` and `readers` could drift

**Taken, and the reviewer is right that my comment claimed more than the spec delivered.** The message is now built from the dispatch table itself rather than from the exported list, so the two *cannot* disagree — and the new spec checks both directions by parsing the words out of the message and comparing the sets, rather than only iterating one of them.

The header comment has been rewritten to say what is actually true: the parser never reads `DECLARATIONS`; it exists for anything outside the parser that wants the list.

### Non-blocking — well-formed items silently swallowed

**Taken rather than accepted.** The reviewer judged this non-blocking as an extension of the enum's pre-existing behaviour, and was careful to say the stated rationale oversells it. That is the part I want to respond to: the comment said *"an author owed three problems is owed all three"* while the code read **past** everything up to the next comma. Either the comment or the code had to go, and the code was the one that was wrong.

`:x [oak, Zeta silver]` kept only `oak`; it keeps `oak, silver` now, with one diagnostic. `:remembers [a: 0, b c: 1]` kept only `a`; it keeps `a, c`. `skipToSeparator` is deleted — the loops guarantee progress past the item they could not read and then carry on, which is what reading *on* means.

**The enum still drops options after a bad one**, and I have not changed that: brace-structured garbage needs a brace-aware skip, and `enum Ward { 4 { message :x } }` is exactly why `recoverInBraces` exists. It is stated in the commit rather than claimed otherwise, and the comment no longer says the list behaves "as the enum does", because it no longer does — it behaves better.

### On the parts that cleared

Recording these because a clean result from ~13,000 generated cases plus three differential oracles is worth more than my own confidence was:

- No input found where the parser needs to know *where* a refused character was rather than merely that one occurred — which is the claim I asked to have challenged when I chose `afterRefusal` over a `bad` token kind.
- No dropped, duplicated or stale `missingComma` across the three loops.
- The deliberately-changed assertion judged right **by execution of the adjacent combinations**, not by reading my justification. That is the check I most wanted, having got exactly this wrong in #58.

### Receipts

Both at `1896bac`. The `e706bfa` receipts are replaced.

```
npm run gate
```
(exit 0 — 811 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 1896bac)
```

No blocking findings outstanding. Nothing fenced. The spec file is untouched.

---

ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:551` (and the identical pattern in `remembers()` at line 685)

The new recovery's `if (this.done) continue;` lets every enclosing list/`:remembers` independently re-report "never closed" for the same exhausted file, so one bad token in a nested, unclosed list is reported as unclosed once per level of nesting around it.

`parseProperty(':x [oak,[Zeta')` — one bad token in a list nested one level deep, never closed — produces `` `Zeta`, which starts with a capital is not a value. `` followed by "This list is never closed." **twice**, both at the identical zero-width span (`k.sprout:1:14`). It scales with depth: `:x [oak, [oak2, [Zeta` (3 levels) repeats the message 3 times; `:x [oak, [oak2, [oak3, [Zeta` (4 levels), 4 times. `:remembers [a: [oak,[Zeta` adds a further "This `:remembers` is never closed." on top of its own doubled "This list is never closed." I checked this is new at this head: I rebuilt `e706bfa` and ran the identical inputs against it — one diagnostic each (just the "Zeta" one), no "never closed" at all through this path. Across a fresh ~20,000-case adversarial run (mutation + random, none of it reused from last round), 370 inputs (1.8%) tripped a "no diagnostic repeats with the same message and span" check I added this round; every single one was this exact message.

Root cause: when the failed element/entry read leaves the parser at `this.done`, execution `continue`s to the top of *this* loop, whose own `this.done` branch reports "never closed" again — even when the failed read was itself a nested `listLiteral()`/`remembers()` call that already reported its own "never closed" for that same exhaustion before returning `null`. The deleted `skipToSeparator`-based code returned directly in this branch (`if (this.separator(']') === 'end' && !this.at('punct', ']')) return null;`) rather than looping back through the top-level check, which is why it didn't double up — it just said nothing further past the nested diagnostic, for this specific case.

Fix: change `if (this.done) continue;` to `if (this.done) return null;` at both sites. Whatever the failed nested read already reported stands as the full explanation; nothing further needs saying once the file has genuinely run out.

---

**Review**: 1 blocking (new), the three prior findings confirmed fixed by re-execution, not just by re-reading the diff. Gate at 1896bac: green, exit 0, last line `✓ corpus: every good archive passes, every bad one fails as expected` (811 passed | 7 skipped in sprout, 40 in cli) — matches the PR's receipts; e2e not re-run (diff still confined to `sprout/lang/src`, author's receipt stands). The `unclosed`-span fix checks out on 1,024 freshly generated decl-keyword-recovery cases, not just the one repro; the `readers.keys()` message and the two "keeps the well-formed neighbor" fixes check out directly. New finding: nested unreadable-and-unclosed list/`:remembers` values report "never closed" once per enclosing nesting level instead of once, confirmed absent at the prior head and present in ~1.8% of a fresh adversarial run — one line per site to fix, in the exact new control flow flagged as highest-risk.

---

*Note on method from the reviewer, relayed verbatim because it is worth more than the bug:*

> my *first* look at this new control flow used a differential oracle whose GOOD/BAD token pools didn't include `[`/`]` as elements — it came back clean (0/16,000) because it never generated a nested list, not because there was nothing to find. I only found the bug once I went back to hand-adversarial probing specifically aimed at "what does a failed nested call leave behind," then confirmed and scoped it by execution (including rebuilding the prior head to check it wasn't pre-existing). Flagging this so the pattern is legible for next time: a clean oracle run is only as good as its input space, and this file has now taught that lesson to two different fuzzing setups in a row.

*Posted on behalf of the `pr-review` agent, which re-reviewed at `1896bac`.*

---

ericeslinger:
## Answering the re-review — and a gate failure of my own

Fixed in `afebfc1`, with a correction in `4e913d0`. Taking the second one first, because it is mine and it is worse than the bug.

### I committed and pushed with a red gate

`afebfc1` failed `typecheck:spec`. I ran the gate, it exited 1, and I committed anyway because I read the test counts — 818 passed — and not the exit code. CLAUDE.md is not ambiguous: *the gate runs before every commit, green or the commit does not happen.*

The defect was in the generator I added this round: a helper took `fn = parseProperty` as a default, which types the parameter as *that* signature, so passing `parseRemembers` to it does not typecheck. **Vitest strips types, so every test passed.** That is precisely the failure `typecheck:spec` exists to catch, described in CLAUDE.md in those words, and I walked straight into it and then ignored the tool that caught it.

`4e913d0` is the one-line fix, as its own commit rather than an amend, so the record shows it happened. Gate and e2e are green there.

### Blocking — "never closed" once per level of nesting

**Fixed exactly as suggested**, `continue` → `return null` at both sites.

```
:x [oak,[Zeta                before                        after
                              `Zeta` … is not a value.      `Zeta` … is not a value.
                              This list is never closed.    This list is never closed.
                              This list is never closed.

:x [oak, [oak2, [Zeta         … the same message ×3         … ×1
```

And the cases that must keep theirs still do: `:x [oak`, `:x [oak, Zeta`, `:remembers [a: 0` each say it once.

This is **the fourth round on this file, and the fourth time a fix left the input adjacent to the one it fixed broken** — this time in the control flow I had explicitly flagged to the reviewer as the highest-risk part of the previous commit. I do not think that is bad luck any more; it is what this kind of code does.

### What I did about the method note

The reviewer's note is worth more than the finding, so I have acted on it rather than only relaying it:

> a clean oracle run is only as good as its input space, and this file has now taught that lesson to two different fuzzing setups in a row

Their first oracle over this control flow returned 0 failures from 16,000 cases **because its token pool had no brackets**, so it never generated a nested list — the shape every one of the four rounds of bugs here has turned on.

So the suite now keeps a small **deterministic** generator with brackets in the pool, over all three entry points, asserting the invariants the bugs actually broke rather than any particular input:

- it never throws (the `RangeError` class),
- it never says one thing twice in one place (this round's class — the exact check that found this bug),
- the same source gives the same answer twice.

Fixed seed, ~600 cases per reader, so it is reproducible and costs the suite almost nothing. It would have caught this round's bug, and it is the first thing in this file that is looking for a *class* rather than a list of inputs.

### Receipts

Both at `4e913d0`. The `1896bac` receipts are replaced.

```
npm run gate
```
(exit 0 — 818 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 4e913d0)
```

No blocking findings outstanding. Nothing fenced. The spec file is untouched.

---

ericeslinger:
**Non-blocking** — `sprout/lang/src/parse.spec.ts:952`

`'Zeta'` in the enumeration's `BAD` pool never exercises `listLiteral`/`remembers` recovery at any depth, through either reader — it's read as an attempted list type instead, because `atType()`'s bracket-skipping check doesn't care how many `[` precede a capitalized token.

`` :x [Zeta `` and `` :x [[[Zeta `` etc. all produce "A list type is never closed." (from `typeExpr`), never "is not a value." or "This list is never closed." (from `listLiteral`) — confirmed across all 6 depths × 3 endings × both readers (36 shapes for `'Zeta'` alone), 0 of which reach the fixed code. It's also the only `BAD` member whose "closed with a good neighbour" shape doesn't keep the neighbour (`` :x [Zeta, oak] `` → `declared === null`, "A list type names one element type.") — not a bug, since that message is pre-existing and correct, but it means `'Zeta'` contributes nothing to the "keeps a good neighbour" coverage either. The other five members are genuine and well-covered.

Fix, if you want the pool at its stated size: replace `'Zeta'` with something whose first token isn't `kind`-shaped, e.g. a second copy of an already-distinguishing bad token, or restructure `nested`/`shapes` so each level gets a leading lowercase sibling (`oak0, [oak1, [...`), which forces `atType()` to fail at every level the way the actual historical bugs (`:x [oak,[Zeta`, `:x [oak, [oak2, [Zeta`) were shaped. Either way, a one-line comment next to `BAD` saying why, matching the one you already left for `'enum'`.

---

**Review**: no blocking findings, 1 non-blocking. Gate at af3b5f6, re-run myself rather than trusting the receipt: green, exit 0, last line `✓ corpus: every good archive passes, every bad one fails as expected` (821 passed | 7 skipped in sprout, 40 in cli), `typecheck:spec` confirmed running and green in the log. The `return null` fix holds up under 8,528 cases specifically constructed to reach and cross the nesting cap, stack two bad tokens in one structure, and break multiple `:remembers` entries at once (94.6% confirmed to actually reach the fixed code) — no throws, no duplicate or dropped "never closed," no nondeterminism — and I can argue by induction over the five mutually-recursive readers that the new early return can never lose a diagnostic. One non-blocking finding: `'Zeta'` in the new enumeration's `BAD` pool never reaches the code the suite says it's covering, at any depth.

---

*The three questions the reviewer was asked, answered:*

**1. A fifth round of adjacent breakage.** Not found.

> I looked hard, past what your own enumeration covers (deeper than 6, crossing the nesting cap, two bad tokens in one structure, multiple broken `:remembers` entries at once) — see the 8,528-case run above. I'm reporting "looked and didn't find one" rather than manufacturing something to satisfy the prior.

**2. Whether the enumeration is vacuous.** Partially — the finding above. Of note:

> This is the same *class* of thing `'enum'` was, just a different mechanism — not secretly legal, secretly *routed elsewhere*.

**3. `return null` versus `continue`: no loss.**

> every `null` that `literal()`, `listLiteral()`, `typeExpr()`, `propertyBody()`, and `rememberedProperty()` can return is preceded by a `diagnostics.refuse(...)` call — either directly, or by delegating to another of those five whose null obeys the same rule. It's a closed set […] so it holds by induction on nesting depth, base case being the branches that refuse directly. Reaching `if (this.done) return null` requires `element === null`, which by that induction means something was already said. And since `this.done` means the token stream is actually exhausted, there's nothing after that point to abandon either.

---

*Relayed verbatim, because it is the third time in a row this file has taught this lesson, and this time to the reviewer:*

> my own first pass at a deeper fuzzer this round made the same mistake in miniature — I built nested shapes as `` :x oak, [oak0, [Zeta `` without an enclosing bracket, and `parseProperty` just read `oak` as the whole value and silently ignored everything after the first comma (nothing requires it to consume the rest of the source). It came back 0/7,574 failures for the same reason your first oracle came back 0/16,000 — not because there was nothing wrong, but because it wasn't looking. I added a self-check (does the run's own diagnostic vocabulary show it actually reached the value path) before trusting the next version, which is what caught it and what the 94.6% figure above comes from.

*Posted on behalf of the `pr-review` agent, which re-reviewed at `af3b5f6` in a fresh worktree outside the shared checkout. `gh` is not installed in this environment, so it returned its findings rather than posting them itself.*

---

ericeslinger:
## Answering the third review

Fixed in `11eb88b`. One non-blocking finding, and it is the one I most wanted found.

### Non-blocking — the anti-vacuity net was itself vacuous

**Taken, by the second route the reviewer offered.** The enumeration I added last round exists *because* two fuzzing setups had come back clean while never generating the shape that has produced a bug in every round on this file. It then did the same thing itself, for a sixth of its own table.

`atType()` skips past every leading `[` before it decides whether what is underneath looks like a type. So `:x [[[Zeta` is an attempted list *type* and goes to `typeExpr()` — which this PR never touches:

```
":x [Zeta"                   A list type is never closed.
":x [[[Zeta"                 A list type is never closed.
":x [oak0, Zeta"             `Zeta`, … is not a value. | This list is never closed.
":x [oak0, [oak1, Zeta"      `Zeta`, … is not a value. | This list is never closed.
```

36 of the 216 shapes — every `'Zeta'` one, at all six depths, through both readers — were asserting invariants about code they never ran.

I took the restructure rather than swapping the token out, because it is the shape correctness argument wants: every level now opens with a lower-case element (`[oak0, [oak1, …`), which is what forces the brackets to be *values* at every level, and it is exactly what the four historical bugs looked like (`:x [oak,[Zeta`, `:x [oak, [oak2, [Zeta`). `'Zeta'` stays in the pool and now earns its place.

**And the assertion is no longer a matter of my word.** A new first test asserts the coverage rather than assuming it — no shape may produce a diagnostic from the type path's vocabulary, every shape must say *something*, and the count of shapes reached must equal the size of the table:

```ts
expect(both.some((m) => m.includes(TYPE_PATH)), `${what}: ${text} went to the type path`).toBe(false);
expect(both.length, `${what}: ${text} said nothing`).toBeGreaterThan(0);
…
expect(reached).toBe(6 * BAD.length * 3 * 2);
```

That test fails on the version of the enumeration I shipped last round. It is the first thing in this file that checks a *suite* rather than the parser, and on the evidence of three rounds that is a check this file needs. The comment next to `BAD` now says why every level has a lower-case head, beside the one already there for `'enum'`.

### Two test-only artifacts fell out of writing it

Neither is a parser bug; both were mine, and both would have been read as parser bugs by the next person:

- I shared one `Diagnostics` across the property and `remembers` reads, which made the first read's diagnostic look like a repeat from the second. Each read gets its own now.
- `:x [enum]` is correctly silent — nothing reserves an option's name, so that is a legal list holding one option. `'enum'` was already out of the pool for this reason; the comment now says so rather than leaving it to be rediscovered.

### On the parts that cleared

Recording these because the reviewer argued them rather than merely failing to find a counterexample:

- **No fifth round.** Looked for past the enumeration's own reach — deeper than 6, across the nesting cap, two bad tokens in one structure, several broken `:remembers` entries at once — and reported *not found* rather than manufacturing one. That is the right answer to give and the harder one to give.
- **`return null` loses nothing**, by induction over the five mutually-recursive readers: every `null` any of them returns is preceded by a refusal, directly or by delegation, and the set is closed. I had this as a belief; it is an argument now.
- 8,528 generated cases, 94.6% *confirmed by their own diagnostic vocabulary* to reach the fixed code — the self-check the reviewer added after their own first fuzzer came back 0/7,574 for want of an enclosing bracket. Third setup in a row to need it.

### Receipts

Both at `11eb88b`. The `4e913d0` receipts are replaced.

```
npm run gate
```
(exit 0 — 822 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 11eb88b)
```

No blocking findings outstanding, none this round. Nothing fenced. The spec file is untouched.

---

PR #61, opened 2026-09-21
B08: typed bindings, exactly as the table

### LINKED ISSUE #8: B08 · Typed bindings, exactly as the table
**Backlog item B08** · Phase 1 — The type system · size M

## What

Typed bindings, exactly as the table: `self`, `actor` as the visitor kind, `here`, `mover`, roles by kind or value type or the bare object type, set roles, `from`-narrowed roles, `each` and `{for}` variables by kind filter, handler senders as objects, handler values by declaration, `was`, `elapsed`, `let`

## Spec

Implements: **Properties › Where types come from** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#3 (B03), #5 (B05), #6 (B06), #7 (B07)

## Blocks

#9 (B09), #23 (B23)

## Working on this

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B08.

ericeslinger:
Closes #8

*Every binding is typed where it enters scope. There is no unknown receiver anywhere in the language.* `bindings.ts` is where that claim is kept: one constructor per row of **Properties › Where types come from**, so the file and the table can be read side by side, and a `Scope` that says what is in reach and refuses a second thing answering to one name.

Nothing calls it yet — there are no bodies to check until B09, no verbs until B23, no handlers until B32. This is the table those three read.

## A binding's type is not a property's type

A property holds a value and there is no object-valued property; a binding may be an **object**, or a **set** of them, neither of which can be written down. So `BindingType` is three arms rather than a `SproutType`:

```ts
| { binds: 'value';  type: HeldType }
| { binds: 'object'; kind: KindRef | null }   // null is the object type
| { binds: 'set';    kind: KindRef | null }
```

The object arm carrying null is the whole of what the object type means: *the type of a binding whose kind the compiler cannot know.* It is one representation rather than two, so nothing has to remember that an unkinded object and the object type are the same thing.

`HeldType` is `SproutType` minus the object arm. It is called that and not `ValueType` for the reason `SproutType` is not called that either — the previous language's extension interface still exports the name, and B45 takes it back. The build told me so: the first version was called `ValueType` and `tsc` found the collision.

## The table, row by row

| row | constructor |
| --- | --- |
| `self` — the composed kind | `selfBinding(kind, at)`, and the only one with `writable: true` |
| `actor` — the world's visitor kind | `actorBinding(visitor, at)` |
| `here`, `mover` — object | `hereBinding`, `moverBinding` |
| a role — the kind or value type the verb declares, object if none | `roleBinding(name, declared, narrowing, at, …)` |
| a set role — as above, as a set | `setRoleBinding(name, kind, at)` |
| a role narrowed by `from` | `roleBinding` with a `RoleNarrowing` |
| an `each` / `{for}` variable — its kind filter, or object | `loopBinding(name, filter, at, inProse)` |
| a `{for}` variable over a list | `forElementBinding(name, element, at)` |
| `each … of` a set role | `setMemberBinding(name, kind, at, inProse)` |
| a `let` binding — the expression it names, exactly | `letBinding(name, type, at)` |
| a handler's sender — object; its value — the declaration | `handlerParameters(message, written, at, …)` |
| a hook's previous value — the property that changed | `wasBinding(name, property, at, …)` |
| `elapsed` — integer | `elapsedBinding(name, at)` |

Every binding carries a **span** rather than an optional one: the name as written, or the member that brought it into scope where the engine supplies the name. A diagnostic about a binding has to point somewhere, and there is always something to point at.

## What a kind is, and is not

B12 declares kinds and B19 composes them. What *typing* asks of one is `KindRef` — identity, the closure of everything it composes, and what it declares — and that is all this file knows:

```ts
export function composesKind(kind: KindRef, target: KindRef): boolean {
  return kind.composes.has(kindName(target));
}
```

Matching is **nominal and by composition, never structural**, so the closure is a set and asking is one lookup rather than a walk. `printers_shop.Vessel` is not `other_world.Vessel` for looking alike — there is a spec for exactly that, because a slot that accepted anything shaped like a container would readmit the confusion that makes two kinds declaring `:open` a collision in the first place.

## Three things that are binding rules rather than message rules

- **A handler's two parameters, positional.** `_` leaves one unnamed and either may be omitted, so what is passed in is what was written and the positions give it meaning. The spec's own four handlers — `(from, value)`, `(from)`, none, `(_, value)` — are a spec.
- **A hook's `was`**, typed by the property that changed.
- **The six engine messages**, which bind what they name rather than a sender and a value: `:entered (item, from)`, `:left (item, to)`, `:moved (from, to)`, `:spawned (from)`, `:tick (elapsed)`, `:woke (elapsed)`. The *reserved-name refusal* for those six is **not** here — *Reserved names* is B23's spec section, and this PR adds only the binding table.

## Shadowing, and the one re-binding that is not shadowing

`introduce` refuses a name anything already answers to, looking outward, and leaves the scope as it was — so the first name still means what it did. The refusal names what the first one was: *"`tools` already names this verb's role here."*

`narrowing` is the exception, and it is the whole of how anything of object type is read:

```ts
narrowing(binding: ObjectBinding, kind: KindRef): Scope
```

It takes the **binding**, not the name, so it cannot be called for a name nothing answers to; and it takes an `ObjectBinding`, so it cannot be called for a value. `isObjectBinding` is the predicate that gets you one. It does not introduce a second thing answering to the name — it says more about the one already there — and it lasts for the branch and no longer.

Neither `lookup` nor `names` recurses. A scope chain is as deep as the blocks a body nests, and a `RangeError` is not a diagnostic — see below.

## Three exact return types in `types.ts`

`integer()` declared `SproutType`, so an integer's range did not survive being built: every caller wanting a value type had to widen and narrow again to get back what it had just asked for. It returns its own arm now, and `BOOLEAN`/`STRING`/`OBJECT` are `as const satisfies SproutType` for the same reason. No behaviour changes; four call sites in this PR stop needing a cast.

## Method

`parse.ts` taught this repository three times in a row that a green suite is worth what it would catch. So before trusting this one, **eighteen deliberate defects were introduced into `bindings.ts` and run against it. All eighteen failed it.**

Among them: `self` not writable; kinds matching by bare name; shadowing caught only in the same scope; a narrowing that leaks past its branch; `lookup` preferring the outer scope to the nearer one; `names()` repeating a narrowed name; a value bound on a message that carries none; an integer role forgetting the property's range; `elapsed` bound as an object; a set role binding one object; a `from` giving the list rather than its element type.

A nineteenth survived and was **my mistake in writing it** — it only added a comment, so it changed nothing and surviving was correct. The real version of it was caught. That check is not in the suite; it is why I believe the suite.

The suite itself builds its fixtures the way the other suites do — real source through `parseDeclarations`/`parseProperty` and the B05–B07 tables — so a `ResolvedProperty` here is one the compiler would actually make. Kinds have no syntax yet, so a `KindRef` is built by hand, which is exactly the seam this file is about.

## What the review found, and what it was

Two non-blocking findings, both taken in `d3ee947`, both reproduced before being touched.

**`Scope.lookup` and `Scope.names` threw `RangeError` on a deep chain** — `names` from about eight thousand levels, `lookup` from about twenty — out of a file whose whole job is producing diagnostics. The first draft of both was **iterative**; I rewrote them as recursion *in this same PR* to satisfy `@typescript-eslint/no-this-alias`, and turned a lint fix into a crash. Nobody reported that lint error — the gate did, and I took the first shape that made it green without asking what else it changed. Both read outward from `this.parent` now, which does not alias `this`, so nothing recurses and the rule stays satisfied.

I did not take the other half of the suggestion, counting scope depth against the host's `nesting` cap. The parser counts because it *creates* the nesting and has a written bracket to point at; a `Scope` is handed a chain somebody else built and has nothing to name in a refusal. That cap belongs in B09's AST walk. What this layer owes is not to die.

**`setMemberBinding` always reported origin `each`**, where `loopBinding` ten lines above takes an `inProse` flag — so a name bound by a prose `{for x of <set role>}` was called "the loop variable" rather than "the passage's loop variable". Only the wording of a refusal, but an asymmetry B29 would have inherited without noticing.

Three specs added for the two, each checked to **fail against the code it replaces**: `lookup` and `names` through a chain of forty thousand, a shadow still refused from the bottom of one, and the two loop constructors agreeing about the flag so they cannot drift.

## Spec gaps recorded

Four, in the working notes' *Holes in the spec*:

- **What a `symbol` role binds to where the role-player wrote no `from`.** *"Without a `from`, a `symbol` role has no options at all and the phrase never matches — the role-player must say what it can hear."* That describes a phrase that never matches, not a refusal; but the body still has to compile, and with no `from` there is no enum to type the binding by. **Refused, naming what to write.** If the intent is that such a body compiles and is simply unreachable, the binding needs a type the spec does not supply.
- **What an `integer` role binds to with no `from`.** The symbol case is stated and this one is not. Taken as the whole integer range — the reading that makes `from` a narrowing rather than a requirement, since an integer role has values to offer without one where a symbol role has none.
- **Whether a value role may be marked `many`.** *Set roles* is written entirely about objects; *Value roles* never mentions `many`. A set role binds objects here. B23 declares roles and is where it has to be settled.
- **What `$first`, `$last`, `$index` and `$count` are typed as.** *Slots* names all four; the table holds none of them, though it opens by saying every binding is typed where it enters scope. Boolean, boolean, integer, integer is the only reading, but it is inferred rather than written. B29's.

## A spec change I am proposing rather than making

Two rows of the table are **narrower than the prose they summarise**, and the prose is the one implemented. I have not touched the spec:

- **"a role narrowed by `from` | the element type of the property named"** is the symbol case only. *A role-player narrows its own options* says an `integer` role's `from` names *"an integer property whose range bounds it"* — that property's **range**, not an element type — and *"a `from` may also give a literal range, `topic from 1 to 12`"*, which names no property at all. Three cases, one row.
- **"a `{for}` variable | the same"** as an `each` variable holds for `{for x in …}` and `{for x: Kind in …}`. *Slots* also has `{for x of <list>}`, which binds the list's **element type** — a value, where every `each` binds an object, and the one loop `each` deliberately does not have.

Both are recorded in the notes so they survive this PR. Whether the table should gain the rows is yours.

## Receipts

Both at `d3ee947`, the PR head. The `29c5ac6` receipts are replaced.

```
npm run gate
```
(exit 0 — 888 passed | 7 skipped across lang/core/stores, 40 passed in cli; 822 before this PR)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at d3ee947)
```


ericeslinger:
**Non-blocking** — `sprout/lang/src/bindings.ts:535` (`Scope.lookup`) and `:547` (`Scope.names`)

Both recurse over the parent chain with no depth guard or iterative fallback.

Bisected directly: a `Scope` built by repeated `.inner()` resolves `lookup()` correctly through 8,000 levels of nesting, but throws an uncaught `RangeError: Maximum call stack size exceeded` — not a diagnostic — somewhere between 8,000 and 12,000; `names()` shares the shape and fails the same way. Nothing in this PR drives that depth (B09/B12/B19/B23 aren't built yet), so it isn't reachable through any real path today, and it needs an implausible amount of nesting for a "small" world even once something does drive it.

Flagging because the project has already hit this exact class of problem once, in the parser: `sprout-working-notes.md`'s B02 entry says "without a bound, seven thousand opening brackets crash the compiler instead of refusing," and fixed it by counting against a host `nesting` cap rather than recursing unbounded. Worth the same treatment here (or an iterative `lookup`/`names`) before B09 walks an AST and calls `.inner()` once per nested block — otherwise deep nesting will crash the checker instead of refusing it.

---

**Non-blocking** — `sprout/lang/src/bindings.ts:248`

`loopBinding` takes an `inProse` flag so `each pot: Vessel in self` reports origin `'each'` ("the loop variable") and `{for pot: Vessel in self}` reports `'for'` ("the passage's loop variable"), matching the spec's two forms of walking a container. `setMemberBinding` — the parallel constructor for "each … of a set role" — has no such flag and always reports `'each'`, even though *Slots* gives a set role the same two forms (`each tool of tools` in code; `{for x of <set role>}` in prose, line 1149).

Only affects the wording of a shadow-collision diagnostic for a name bound from a prose `{for … of <set role>}` ("the loop variable" instead of "the passage's loop variable"), so no type decision is wrong — but it's an easy asymmetry for whichever issue wires up prose (B29) to inherit without noticing, given `loopBinding` right above it already shows the pattern.

---

**Review**: no blocking findings, 2 non-blocking. Gate at 29c5ac6: green (re-ran `npm ci && npm run gate` at the PR head in a clean worktree — exit 0, test counts and last line match the receipt exactly, and `typecheck:spec` produced no errors across every workspace). Checked the table row by row against the spec (523, 735, 755, 786, 959, 1113); all fourteen rows match, including the two rows the working notes flag as narrower than their own prose, correctly implemented from the fuller prose rather than the compressed row and disclosed as such. Fuzzed `Scope` (~18,000 randomized operations against a reference model) and swept `roleBinding`'s `from` narrowing and `handlerParameters` combinatorially against the built code; found no wrong-type or wrong-refusal case. The four holes in the working notes are genuine — no sentence in the spec settles a `symbol` role with no `from`, or whether a value role may be `many` — and the "refuse rather than compile-as-unreachable" call is the defensible reading given the project's own no-`any`-at-a-boundary rule. `KindRef` and the `types.ts` return-type tightening are justified, not scope creep: `HeldType` excludes the object arm, and the tightened returns are what let `bindings.ts` assign into it without a cast.

---

*On the judgement call the review was asked about specifically — refusing a `symbol` role with no `from`, where the spec arguably says the phrase merely never matches:*

> I think it's the right call and effectively the only implementable one. Treating the body as compiling-but-unreachable would require typing `topic` as something with no real `SproutType` behind it (there's no enum to check `topic == :toll` against), which is exactly what "no `any`/`unknown` at a language boundary" rules out.

*On the seam:*

> `KindRef` matches the spec's own vocabulary precisely: "a kind's identity is its library and its name" (line 332) and "an object's properties are one flat map" (line 289, *Properties merge; they never shadow*) — by the time a `KindRef` exists, B19 has already resolved merge collisions, so a flat map is the right shape for what typing consumes. It asks nothing of B12/B19 beyond identity, closure-as-a-set, and declared members.

*Posted on behalf of the `pr-review` agent, which reviewed at `29c5ac6` in a scratch worktree outside the shared checkout, removed afterwards. `gh` is not installed in this environment, so it returned its findings rather than posting them itself.*

---

ericeslinger:
## Answering the review

Both taken, in `d3ee947`. Both reproduced by execution before being touched.

### Non-blocking — a deep scope chain throws instead of saying something

**Fixed, iteratively.** The reviewer's bisection is right, and I narrowed it further against the built module:

```
depth   1000  lookup: ok                 names: ok
depth   4000  lookup: ok                 names: ok
depth   8000  lookup: ok                 names: THREW RangeError
depth  12000  lookup: ok                 names: THREW RangeError
depth  20000  lookup: THREW RangeError   names: THREW RangeError
```

**Worth naming what this was: the first draft of both was iterative.** I rewrote them as recursion *in this same PR*, to satisfy `@typescript-eslint/no-this-alias`, and turned a lint fix into a crash — in a file whose whole job is producing diagnostics. Nobody reported the lint error; the gate did, and I took the first shape that made it green without asking what else it changed.

Both now read outward from `this.parent`, which is not aliasing `this`, so the rule stays satisfied and nothing recurses:

```ts
const own = this.bindings.get(name);
if (own !== undefined) return own;
for (let scope = this.parent; scope !== null; scope = scope.parent) { … }
```

I did **not** take the other half of the suggestion — counting scope depth against the host's `nesting` cap. The parser counts because it is the thing that *creates* the nesting and has a written bracket to point at; a `Scope` is handed a chain somebody else built and has nothing to name in a refusal. The cap belongs where nesting is counted, which is B09's AST walk. What this layer owes is not to die, and now it doesn't.

Three specs, at a depth of forty thousand — far past anything a body nests, and a few milliseconds: `lookup` finds a name and returns null for one nothing answers to; `names` lists what is in reach; and a shadow is still refused from the bottom of the chain. Each was checked to **fail against the recursive version**, not merely to pass against the new one.

### Non-blocking — `setMemberBinding` lost the flag its neighbour has

**Taken.** Confirmed first:

```
loopBinding      in code : each      in prose: for
setMemberBinding always  : each
```

A set role has the same two forms a container does — `each tool of tools` in a body, `{for x of <set role>}` in a passage — so it takes the same `inProse` flag, and a refusal about the name calls it what the author called it. The reviewer's framing is the part that matters: *"an easy asymmetry for whichever issue wires up prose (B29) to inherit without noticing"* — the two constructors sit ten lines apart, and one of them silently answered for both forms. There is a spec now that asserts both constructors agree about the flag, so they cannot drift.

### On the parts that cleared

Recording these, because they were argued rather than assumed:

- The table was checked **row by row against the spec's line numbers**, including the two rows the working notes flag as narrower than their own prose — confirmed implemented from the fuller prose and disclosed rather than silently resolved.
- ~18,000 randomised `Scope` operations **against a reference model**, plus combinatorial sweeps of `roleBinding`'s `from` narrowing (4 × 11) and `handlerParameters` (3 × 9). No wrong-type or wrong-refusal case. The robustness gap above is what that testing turned up instead, which is the honest outcome.
- The four holes were checked against *Value roles*, *A role-player narrows its own options* and *Set roles* **in full**, and judged genuine. On the call I flagged as least certain, the reviewer's argument is better than mine was: typing an unreachable `symbol` role would need something with no real `SproutType` behind it, which is the thing the no-`any`-at-a-boundary rule exists to forbid.
- `KindRef` checked against the spec's own vocabulary — "a kind's identity is its library and its name" (332) and the flat property map that *Properties merge; they never shadow* (289) leaves behind once B19 has resolved collisions.

### Receipts

Both at `d3ee947`. The `29c5ac6` receipts are replaced.

```
npm run gate
```
(exit 0 — 888 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at d3ee947)
```

No blocking findings outstanding, none this round. Nothing fenced. The spec file is untouched.

---

Issue #24, opened 2026-09-21
B24 · `as <role> for <verb>` and `as actor for <verb>`


ericeslinger:
**Backlog item B24** · Phase 4 — Verbs, grammar and prose · size L

## What

`as <role> for <verb>` and `as actor for <verb>`: the consent pass and the effect pass in declared order, first refusal as the whole output, roles composing, `from` narrowing, `nothing_happens` when nothing was said to the actor

## Spec

Implements: **Verbs › Playing a role, The actor's own part, The two passes, Roles compose, A role-player narrows** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#22 (B22), #23 (B23)

## Blocks

#25 (B25), #26 (B26), #30 (B30), #34 (B34), #48 (B48)

## Working on this

1. Do not start until every issue under *Depends on* is closed. Issue numbers equal backlog numbers: item B*nn* is issue #*nn*, and lower numbers come first.
2. Claim it: assign yourself, or comment that you are taking it, before starting.
3. Read the spec sections above and the relevant *Decisions* in the working notes before writing code. Existing code under `sprout/` and `cli/` describes the previous language and is not authoritative; replace it where it disagrees with the spec.
4. Open a PR that references this issue. Where the spec is silent or unclear, say so in the PR and add a line to the working notes' *Holes in the spec* — do not invent quietly.
5. When this closes, add the `ready` label to any issue under *Blocks* whose other dependencies are also closed.

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B24.

ericeslinger:
A note from B09 (#62), left here because this is where it will matter rather than where it was found.

[...]

PR #62, opened 2026-09-21
B09: expression checking, and the expressions to check

### LINKED ISSUE #9: B09 · Expression checking
**Backlog item B09** · Phase 1 — The type system · size M

## What

Expression checking: the checker table in full, `is()` narrowing as the only read through the object type, `holds`, `count`, `includes`, memory `recall`/`remember`/`adjust`, literal range checks, no truthiness or coercion

## Spec

Implements: **Properties › What the compiler checks** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#8 (B08)

## Blocks

#11 (B11), #29 (B29), #50 (B50)

## Working on this

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B09.

ericeslinger:
Closes #9

The checker table in full — plus the grammar it needs, because an expression cannot be checked before it can be read, and no other backlog item reads one.

## The parser

Precedence climbing over **one table**, in the same `Parser` class as the declarations, so there is one answer to "what token am I at" — which is what #59 was opened about.

```ts
const BINARY_PRECEDENCE: ReadonlyMap<string, number> = new Map([
  ['||', 1], ['&&', 2], ['==', 3], ['!=', 3],
  ['<', 4], ['<=', 4], ['>', 4], ['>=', 4], ['+', 5], ['-', 5],
]);
```

One loop reads every level. There is a spec listing **every expression the spec writes**, with the shape it must parse to, so the grammar cannot drift from what the language is written in:

| written | reads as |
| --- | --- |
| `p != self && chance(4)` | `((p != self) && chance(4))` |
| `!self.get(:inked)` | `(!self.get(:inked))` |
| `from.get(:opens).includes(self.get(:ward))` | a chain of two readings |
| `to.is(sprout.Container)` | a kind, not a member — the capital after the dot decides |

Parentheses, call arguments and stacked prefix operators all count against the host's `nesting` cap, against the **same** `depth`; a chain of operators at one level is read by the loop.

## The checker

Three rules shape the whole file.

**There is no truthiness and no coercion.** `&&` takes booleans, `<` takes integers, nothing is converted on the way in. A condition that is an integer is a refusal, not a zero-test.

**Only `self` writes `self`** — through the binding B08 marks `writable`, not by comparing against the name. And nothing but `remember` writes memory: the four that write draw the held-versus-remembered line `get` draws, each pointing at the reading that keeps what the author meant.

**`is()` is the only read through the object type.** Every other reading of an unnarrowed object refuses and says to narrow it first. `narrowingOf` hands back the `ObjectBinding` and the `KindRef` for `Scope.narrowing` to apply to the branch.

`typeOf` walks the unbounded spines **iteratively**:

> A tree is as deep as its longest chain of operators, and `a + a + a + …` has no bracket in it to count against the host's nesting cap — so a recursive walk would answer a long enough expression with a stack overflow rather than with a diagnostic.

Two hundred thousand terms type without throwing. This is the same class the review of #61 found in `Scope`; I went looking for it here before writing the walk rather than after.

## The nesting cap, and what it took to get right

Two defects with one root — **a construct the cap refuses does not clean up after itself** — one shipped, one new here.

`:x [[[[…` reported the cap **1,992 times**, once per bracket. And a refused bracket is *consumed* without opening a level, so its closer is left in the stream, where the loop reading around it takes the closer for its own and ends early:

```
:remembers [a: <nine deep>, b: 3]     before   kept  a        (b gone, nothing said)
                                      after    kept  a, b
```

Both are fixed at the sites that open a bracket, with one helper. The reviewer recommended a separate issue for the older half; I overruled that, because the fix is smaller than the failure looked and leaving one half would give one shape two behaviours in one parser — the condition #59 was opened to remove. The reviewer agreed on re-review.

**It took three goes, and each wrong turn was caught by measurement rather than by reading.**

1. Stopping the construct at the first too-deep item said the depth once and silently lost every *other* thing the author had got wrong.
2. Skipping to the closer ran past the end of a declaration when the closer was never written, swallowing a whole second `message`.
3. Guarding that with "stop at a declaration keyword" **reintroduced the stray closer the helper exists to prevent** — nothing reserves `message`, so `[message foo]` is a list of two things, and a keyword-lookalike inside an abandoned bracket trapped the skip and lost a sibling entry with nothing said.

The helper now **looks for its closer before consuming anything**: found, it takes exactly that construct; not found, it takes nothing, which is what an unclosed bracket already has recovery for. There is no guard left to get wrong. Specs pin each of the three failures.

`typeExpr` has the same gap one level of self-recursion away — pre-existing on `main`, not reachable through real source until B12 and B19 read property bodies. **#63** carries it, with the reachability analysis and the acceptance test.

## A quadratic lexer, found by that fix

`peek` buffers as far as it is asked, and `next()` drained the buffer with `shift()` — quadratic in its length. Looking ahead over a fifty-thousand-deep expression took one spec **from a millisecond to fifty-two seconds**. `next()` walks a cursor instead.

A change to B01's file, so it has specs in B01's file — six of them. **The timing guard had to be sized by measurement:** the first draft peeked fifty thousand tokens, and a mutant putting `shift()` back *survived*, because `shift()` drains fifty thousand in 222ms against a five-second timeout. Four hundred thousand takes nineteen seconds. The spec uses four hundred thousand and says why, with the count and the timeout both marked not to be touched.

The reviewer verified the fix by reverting only `lexer.ts` on top of this round's parser and watching the same test take 62s and time out — because "I made it fast" and "I made the guard stop noticing" look identical from outside.

## Seams

- **`KindRef` gains `contains`**, which B13 fills. `x.count` and `x.holds(y)` are the two readings that need a container, and containment is a declaration.
- **`KindLookup`** is the two questions "a kind in scope" means, mirroring `EnumTable`. B12/B19/B21 implement it.
- **`chance` and `random` are not read.** The backlog puts both words in B33. The parser reads the shape of a call with no receiver so the refusal names the word; `FREE_CALLS` is the empty table B33 fills. **This becomes wrong rather than merely incomplete the moment B24 lands** — B33 depends on B34 depends on B24, so B24 comes first, and it is what makes `if (p != self && chance(4))` writable. Recorded in the working notes and on #24, where it will be found when it matters.

## Left open, deliberately

- **#63** — `typeExpr` leaves a stray closer when the cap refuses a list type. Pre-existing, not reachable end to end today, and in exactly the recovery machinery where two of my own fixes in this PR introduced defects. Adding one more unforced change to it at the end of a three-round review is how a fourth round happens.
- **#64** — a host `nesting` cap in the low thousands overflows the stack instead of refusing, against a file whose docblock says nothing is thrown. **This one arrived with the expression reader in this PR.** The fix is either an iterative expression reader or refusing a cap the parser cannot honour — the second being cheaper but meaning the language tells the host a number, which Limits says it never does. That choice is not one to make quietly at the end of a PR.

## Method

**Twenty-five deliberate defects** before opening; five survived and are covered. **Fifteen more** across three review rounds, over every fix. Two of my mutants were bad (one only added a comment; one had two identical call sites), two more survived correctly and are reported as such, and one "failure" was **my spec, not the parser** — the tree-rendering helper recursed and overflowed before the code under test did.

The reviews' own independent checks: precedence against a separately written reference (169 operator pairs, 2,197 triples); `depth` returning to baseline across ~15 hand-built interleavings of refused and accepted siblings; the 1,992-duplicate regression reproduced exactly at base; and a structural argument that the bracket skip cannot be quadratic, since every caller advances monotonically and any two skips scan disjoint stretches.

## Spec gaps recorded

Three, in the working notes:

- **What binds tighter than what, and whether an expression may be bracketed.** The spec states no precedence and shows no grouping bracket; it never writes a `+` or a binary `-` at all. Every expression it *does* write assumes the conventional order, so that is what is implemented, with `( … )` grouping. The conventional reading rather than a written one.
- **Whether a list may be written inside an expression.** Every list the spec writes is a declaration's default or a restatement of one, and a list changes through `add` and `remove`. So a list literal is not read in an expression, which leaves nothing unreachable. The narrow reading.
- **Whether `get` may read a remembered property, and `recall` an ordinary one.** Each word reads its own kind, and each refusal names the other. The review found I had asked this of `get` and not of the four that write.

And two that read like gaps and are not: two object bindings of **different kinds may be compared** (*Object identity*), and a `set` whose value cannot be decided until the world runs is **not refused** (*"where the compiler can tell"*).

## Receipts

Both at `0babc60`, the PR head.

```
npm run gate
```
(exit 0 — 955 passed | 7 skipped across lang/core/stores, 40 passed in cli; 888 before this PR)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 0babc60)
```


ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:851` (`unary()`)

Prefix-operator nesting escapes the shared depth cap, so real nesting can run up to 8× past the documented limit of 8 with no diagnostic.

`unary()` counts stacked `!`/`-` against `this.caps.nesting` with a counter local to each call, never touching `this.depth` — the field `deeper()` uses and that parens, list literals and call arguments all share. Every bracketed level therefore gets its own fresh 8-deep unary allowance on top of the shared one. Confirmed: `parseExpression('(' + '!'.repeat(8) + 'a' + ')')` — 8 parens, 8 nots — parses with zero refusals (neither counter individually breaches 8). Pushed further, `('(' + '!'.repeat(8)).repeat(8) + 'a' + ')'.repeat(8)` — 81 characters — also parses cleanly, and the resulting tree is 64 deep (8 paren levels × 8 unary each), eight times the spec's "expression and block nesting: 8" (design spec line 1400) and eight times what the code's own section comment promises: "Parentheses, call arguments and stacked prefix operators all count against the host's nesting cap." They don't; only parens and calls do.

Fix direction: route each prefix operator through `this.deeper()`/`this.depth` instead of a separately-budgeted local counter.

---

**Blocking** — `sprout/lang/src/parse.ts:1008` (`argumentList()`)

A balanced-but-over-cap nested construct inside a call argument silently truncates the rest of the argument list, with no diagnostic that anything was dropped.

`argumentList` reuses `listLiteral`'s read-on recovery ("The same loop `listLiteral` uses, for the same reasons," per its own comment). When an argument's own parens are written balanced but one past the cap, the *excess closing* parens — matching the opens that failed `deeper()` and so never opened a level — are left in the stream, and recovery consumes one of them as the call's own closing `)`, ending the whole argument list early. Confirmed: `parseExpression("self.f(((((((((a))))))))), b)")` (9 opens/9 closes around `a`, cap is 8) returns a **non-null** expression shaped `self.f(a)` — the second argument `b` is gone — with exactly one diagnostic ("Nothing here may be nested more than 8 deep.") and the trailing 11 characters (`))))))), b)`) never examined (the returned node's span ends there). `argumentList` is new in this PR, so this isn't inherited — see comment 4 for the sibling that is.

---

**Blocking** — `sprout/lang/src/check.ts:579-596` (`effectCall`)

`set`, `self.adjust`, `add` and `remove` accept a remembered property as if it were an ordinary held one; only `get` (check.ts:451) enforces the held/remembered split the spec draws.

The spec's memory row (*What the compiler checks*) reserves writing a remembered property to `remember` ("written with `remember`... stepped with `adjust`", *Per-actor memory*), and `getCall` refuses `self.get(:p)` on a remembered `:p`, naming `recall`. But the fallthrough `set`/`self.adjust`/`add`/`remove` share (check.ts:579-596) resolves `:p` via `declaredOn` (line 584), which looks it up in `kind.properties` — a map the `KindRef` doc comment says includes "the remembered ones" too — and never checks `property.remembered`. Confirmed against the built checker with a kind shaped exactly like this PR's own `PRINTER` fixture (composes `sprout.Actor`, declares `:visits` as a remembered `0 min 0 max 99`): `self.get(:visits)` is correctly refused ("Read it with `recall`"), but `self.set(:visits, 5)` and `self.adjust(:visits, 1)` both return `true` with zero diagnostics, and `self.add(:seen, :silver)` on a remembered list property does the same. The working notes record the analogous `get`/`recall` crossing as a considered hole; this one isn't recorded anywhere, which reads as an oversight rather than a decision.

---

**Non-blocking** — `sprout/lang/src/parse.ts:584` (`listLiteral`), `sprout/lang/src/parse.ts:707` (`remembers`)

The identical defect as comment 2 already exists in `listLiteral`/`remembers`, but it predates this PR — confirmed by running the same input against base commit 6d799da: `parseRemembers(':remembers [a: ' + '['.repeat(9) + 'oak' + ']'.repeat(9) + ', b: 3]')` silently drops `b` on both sides of #62, same single diagnostic, same unexamined trailing `, b: 3]`. Not a regression, so not blocking. But the PR is specifically testing and reasoning about the cap's interaction with this exact recovery loop, and its claim that "reporting once per declaration and then refusing in silence is right" reads as a complete account of it — it isn't: a whole later entry can vanish with zero explanation, not just a duplicate report get deduplicated. That's worse than what the "stop the construct" alternative was rejected for. Recommend a dedicated follow-up issue covering all three call sites (it would also cover comment 2) rather than folding more of it into B09. This also bears on "was fixing `listLiteral` inside B09 defensible": the narrow reporting fix is fine to bundle (it's the same root cause `argumentList` needed fixed anyway), but a real fix for the truncation is bigger than either PR attempted and deserves its own issue.

---

**Non-blocking** — `docs/design/sprout-build-backlog.md` (B24/B33/B34), re: the chance/random deferral in the working notes

The deferral to B33 is honestly recorded and harmless today, since no statement/body grammar exists yet to make an expression reachable from real source. But B33 depends on B34, and B34 depends on B24 — so by the backlog's own graph, B24 (which adds `if`/`permit`/`do`, the first place a real `.sprout` file can hold an expression) is a hard prerequisite of B33 and will necessarily land first. The moment it does, the spec's own worked example — `if (p != self && chance(4)) { act nuzzle (target: p) }` (design spec line 804) — becomes writable in a real world, and the shipped checker will refuse it. At that point this stops being "incomplete" and becomes "wrong." Not a problem in #62, which scopes and records this correctly, but worth a note for whoever picks up B24.

---

**Review**: 3 blocking, 2 non-blocking. Gate at 2836487: green — reran in a clean worktree, exit 0, last line `✓ corpus: every good archive passes, every bad one fails as expected`, 942 passed | 7 skipped (sprout) and 40 passed (cli), matching the PR's receipt exactly. Precedence verified against an independently-written reference (169 operator pairs + 2197 triples, all correct); `typeOf`'s spine walk holds on 50,000-long chains; negative and boundary integer ranges check correctly both ends. The nesting-cap regression claim (1,992 duplicate reports at base, `:x [[[[…`) reproduces exactly. All three blocking findings are in new B09 code: prefix-operator nesting silently escapes the shared depth cap (up to 8× over 8), `argumentList`'s comma-recovery can silently drop trailing arguments past a balanced-but-over-cap nesting, and `set`/`adjust`/`add`/`remove` don't enforce the held-vs-remembered split that `get` does. Solid checker-table coverage otherwise (object identity, lists, containment, walking/value-roles correctly out of scope) and the three recorded spec holes are accurate reads of the spec.

*Posted on behalf of the `pr-review` agent, which reviewed at `2836487` in scratch worktrees outside the shared checkout, both removed afterwards. `gh` is not installed in this environment, so it returned its findings rather than posting them itself.*

---

ericeslinger:
## Answering the review

All five taken, in `ac1ef6f`. Every one reproduced against the built code before anything was touched.

### Blocking — prefix signs escaped the shared depth cap

**Fixed as suggested**, and the reviewer's number is exact:

```
8 x (paren + 8 nots)   before   parsed, treeDepth=64, nothing said
                       after    refused once: "Nothing here may be nested more than 8 deep."
```

Sixty-four deep under a cap of eight. The section comment three lines above already claimed *"Parentheses, call arguments and stacked prefix operators all count against the host's nesting cap"* — I wrote that comment in the same commit as the counter that made it false. Each sign goes through `deeper()` now, against `this.depth`, so the comment is true.

### Blocking — a refused bracket left its closer behind

**Fixed**, and this one is the more interesting of the three.

A bracket the cap refuses is consumed and never opens a level, so its closer stays in the stream — where the loop reading around it takes the closer for its own and ends early. A failed bracket now steps over itself, at all four sites that open one.

**On the specific repro:** `self.f(((((((((a))))))))), b)` is balanced — nine opens including the call's own, nine closes — so `, b)` sits *outside* the expression, and `parseExpression` reading one expression and stopping is correct there. The confounder does not save the finding, though: the defect is real, and the `:remembers` case demonstrates it cleanly.

```
:remembers [a: <nine deep>, b: 3]     before   kept  a        (b gone, nothing said)
                                      after    kept  a, b
self.f(<eight deep>, b)               after    self.f(b), one diagnostic
```

**And it took two goes, because I had two things confused.** What is *inside* a construct the cap abandoned goes with it — the way an enum's options go with an enum that could not be read — and one message is the whole account of it. What comes *after* it is a sibling the author is still owed. My first skip also ran past the end of a declaration when the closer was never written, swallowing a second `message` whole; one of my own specs caught that, and the skip now stops at a declaration keyword.

### Blocking — nothing but `remember` writes memory

**Fixed.** `get` drew the held-versus-remembered line and the four that write did not, so a `set` reached memory and wrote one object's idea of everybody at once:

```
self.get(:visits)          refused  "remembered about each actor, not held by the object"
self.set(:visits, 5)       true     (nothing said)
self.adjust(:visits, 1)    true     (nothing said)
self.add(:seen, :silver)   true     (nothing said)
```

The reviewer's framing is the right one: *"the working notes record the analogous `get`/`recall` crossing as a considered hole; this one isn't recorded anywhere, which reads as an oversight rather than a decision."* That is exactly what it was. I wrote the hole about `get` and `recall`, then did not carry the same question to the other four words.

### Non-blocking — the same defect in `listLiteral` and `remembers`

**Fixed here rather than deferred**, and I want to be explicit about overruling the recommendation.

The suggestion was a dedicated follow-up issue across all three call sites. But the fix turned out to be one helper applied where a bracket is opened, not the larger thing it looked like from the failing behaviour — and the two halves are the same defect. Fixing the new half while leaving the old one would give one shape two behaviours in one parser, which is precisely the condition #59 was opened to remove. So all four sites are fixed together and there is no follow-up issue to open.

The reviewer is right that my earlier account oversold itself. *"A whole later entry can vanish with zero explanation, not just a duplicate report get deduplicated"* — true, and I had described only the deduplication. The PR body now describes both.

### Non-blocking — `chance` becomes wrong, not merely missing, the moment B24 lands

**Taken, and acted on rather than just noted.** The graph argument is the part I had not made: B33 depends on B34 depends on B24, so B24 necessarily lands first, and it is the item that makes `if (p != self && chance(4))` writable in a real world. A note is now in the working notes for whoever picks B24 up, and I have put the same note on #24 rather than leaving it only here, where it would be found by nobody at the moment it matters.

### On the parts that cleared

- Precedence checked against an **independently written reference** — 169 operator pairs and 2,197 triples, all correct. That is the check I most wanted, because the precedence is the conventional reading rather than a written one.
- The spine walk holds on 50,000-long chains, and integer ranges check correctly at both ends including negatives.
- The 1,992-duplicate regression at base reproduces exactly, which is the claim the whole cap change rests on.

### Receipts

Both at `ac1ef6f`. The `2836487` receipts are replaced.

```
npm run gate
```
(exit 0 — 945 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at ac1ef6f)
```

Six mutants over the six fixes, each checked to fail the suites. No blocking findings outstanding. Nothing fenced. The spec file is untouched.

---

ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:234` (`skipBracketed`), specifically the unconditional check at line 241

`skipBracketed` calls `this.atDeclarationKeyword()` on every token it scans, not gated by its own local `depth` the way `recoverInBraces` was fixed to do in #59/#60 for the identical failure — so if an abandoned bracket's content contains a bare `message` or `enum` word followed by anything that isn't `,`/`}`/`]`, the skip stops immediately, the real closers are left in the stream, and the next recovery loop mistakes one of them for its own. This reopens the exact bug class (stray closer → silent sibling loss) this PR exists to close.

Confirmed with `parseRemembers` on:
```
:remembers [a: [[[[[[[[[message foo]]]]]]]]], c: 3]
```
(9 `[`, one over the default cap of 8). Result: `properties` contains only `a` — `message` and `foo` get folded into `a`'s value as stray list elements (with a spurious "A list needs a comma between its elements." diagnostic along the way) — and `c: 3` is gone with **no diagnostic of any kind** naming it. That's worse than the pre-fix duplicate-report bug, which at least said something once.

The same shape at file scope only *looks* fine by accident:
```
message :first with [[[[[[[[[message Foo]]]]]]]]]
message :second
```
`:second` does survive, but only because `file()`'s outer loop restarts a fresh declaration attempt at the trapped "message" token and *that* attempt's own `recover()` call happens to sweep the rest — producing a fabricated "A message needs a name." pointing at content the author never wrote as a declaration. That's the "wrong diagnostic somewhere else" you asked me to look for.

I also checked, and these are all fine (no regression): unbalanced brackets in both directions, a closer of the wrong kind (`)` where `]` was expected), a stray `}` inside an abandoned `[`, balanced `(...)` nested inside an abandoned `[`, and a genuinely-never-closed bracket (runs cleanly to EOF, no throw). Depth bookkeeping itself is also sound. The hole is specifically the keyword-lookalike case.

Note for the fix: a literal copy of `recoverInBraces`'s `depth === 1 && this.atDeclarationKeyword()` guard will not close this by itself — in the repro above, the trap token sits at `skipBracketed`'s own local `depth === 1` already, immediately after the token that failed `deeper()`.

---

**Non-blocking** — `sprout/lang/src/parse.ts:219` (`deeper`, reached from `unary()` at line 897)

Routing prefix signs through the shared `deeper()`/`reportCap()` means its hardcoded remedy — "Take some of the brackets out." — now also fires when the offending construct has no brackets at all.

Confirmed: parsing `'!'.repeat(9) + 'a'` (no brackets anywhere) refuses with "Nothing here may be nested more than 8 deep." and remedy "Take some of the brackets out.", which names something the author didn't write and gives no correct instruction about the nine `!` that are actually there. `reportCap` already takes `remedy` as a parameter — `unary()` could pass its own text (e.g. the pre-fix "Take some of the signs out.") instead of relying on `deeper()`'s hardcoded one.

---

**Non-blocking** — `sprout/lang/src/check.ts:589-596` (`effectCall`)

The new "remembered, not held" refusal shared by `set`/`adjust`/`add`/`remove` always tells the author to switch to `remember`, even when the method was `adjust` — which has its own spec'd memory operation (`x.adjust(:p, n)`, *Per-actor memory*) that preserves the step semantics the author wrote; `remember` overwrites instead.

Confirmed against a `Printer`-shaped kind with `:visits` remembered: `self.adjust(:visits, 1)` refuses with "Write it with `remember`, as in `actor.remember(:visits, …)`.", while `actor.adjust(:visits, 1)` — same property, only the receiver changed from `self` to `actor` — is accepted. An author who follows the printed remedy literally trades "step by 1" for "overwrite to a fixed value," not what they asked for. For `set`/`add`/`remove` the same remedy is accurate (or the closest available, since memory has no `add`/`remove`); only `adjust` needs its own wording.

---

**Review**: 1 blocking, 2 non-blocking. Gate at `ac1ef6f`: green (945 passed | 7 skipped sprout, 40 passed cli, corpus clean) — matches the PR's receipt. Two of the three previously-blocking findings are cleanly fixed: the prefix-sign depth cap (verified with ~15 hand-built adversarial cases — successful and refused siblings interleaved in argument lists, nested calls, standalone parses — depth returns to baseline correctly in every direction, no spurious refusal and no reopened hole) and the memory-write split (`actor.remember`, `actor.adjust` on memory, and ordinary `self.set`/`self.adjust` on a held property of a mixed kind all still work). The third (stray-closer recovery) is only partly fixed, per the blocking finding above.

---

*On the five questions it was asked:*

**Depth bookkeeping** — correct, no leak either direction, stress-tested rather than read:

> refused-then-recovered siblings followed by at-cap siblings in call argument lists, at-cap siblings repeated three times in a row, nested calls two levels deep each at the effective cap… `this.depth` always ends up back at exactly the right value. `unary()`'s `operators.length`-based decrement in `finally` is correct because `operators` is only pushed on a successful `deeper()`.

**On overruling the separate-issue recommendation:**

> I don't disagree with that call; "one shape, one behavior" is the right reasoning and the diff is small enough to review as one thing. But given the blocking finding above, the completeness claim in the PR body ("Both are fixed at all four sites… with one helper") isn't quite true yet — same helper, so the same gap is at all four sites too. I'd fix it here rather than reopen the separate-issue question.

**The `chance`/B24 note** — confirmed present in both the working notes and on #24.

*Posted on behalf of the `pr-review` agent, which re-reviewed at `ac1ef6f` in a scratch worktree outside the shared checkout. `gh` is not installed in this environment, so it returned its findings rather than posting them itself.*

---

ericeslinger:
## Answering the re-review

All three taken, in `982b2eb`. All three reproduced before being touched. And one of them found something else.

### Blocking — my own fix reintroduced the bug it was written to close

**This is the sixth time on this parser that a fix broke the input adjacent to the one it fixed, and the second time the fix broke the very thing it was for.** `skipBracketed` exists to stop a stray closer being taken for somebody else's; its guard left the real closers behind and lost a sibling entry silently — which is worse than the duplicate-report bug it replaced, because that one at least said something.

```
:remembers [a: <nine deep, holding `message foo`>, c: 3]
  before   kept  a        (c gone, and nothing said about it)
  after    kept  a, c

message :first with <the same>          before   ! A message needs a name.   (fabricated)
message :second                         after    (nothing fabricated)
```

The guard existed for a real reason — a closer that was never written took the rest of the file with it, which I had also found the hard way — so the fix had to answer both. **It looks for the closer before consuming anything.** Found: the skip takes exactly that construct. Not found: it takes nothing, which is what an unclosed bracket already has recovery for. There is no guard left to get wrong, which is the only version of this I now trust.

The reviewer's note that copying `recoverInBraces`'s `depth === 1` guard would not have closed it is right, and saved me the obvious wrong fix: the trap token sits at local depth 1 already.

### That fix found a quadratic lexer

Looking ahead fills the lexer's buffer, and `next()` drained it with `shift()` — one token at a time out of an array, which is quadratic in its length. A fifty-thousand-deep expression took one spec **from a millisecond to fifty-two seconds**.

It failed on the vitest timeout, which means the suite already guards this without an assertion, so there is now a comment on that test saying *do not raise the timeout to make this pass*. `next()` walks a cursor instead. This is a change to B01's file and I am flagging it rather than burying it: it is a performance fix with no behaviour change, and the lexer spec plus 949 others hold.

### Non-blocking — two remedies that named the wrong thing

Both taken.

```
!!!!!!!!!a   before  "Take some of the brackets out."   (there are no brackets)
             after   "Take some of the signs out."
```

Routing prefix signs through `deeper()` last round also gave them its remedy. `deeper()` takes one now, and `unary()` passes its own.

And the memory refusal sent `adjust` to `remember`. The reviewer's reasoning is the part worth keeping: *"an author who follows the printed remedy literally trades 'step by 1' for 'overwrite to a fixed value'"* — a remedy that is true about the type and wrong about the meaning. `adjust` has its own memory reading and is pointed at it; `set`, `add` and `remove` keep `remember`, which is the closest thing that exists.

### On the completeness claim

The reviewer was right that the PR body's *"fixed at all four sites with one helper"* was not yet true — same helper, so the same gap was at all four. It is true now, and the body says what the helper actually does rather than what I hoped it did.

### Receipts

Both at `982b2eb`. The `ac1ef6f` receipts are replaced.

```
npm run gate
```
(exit 0 — 949 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 982b2eb)
```

Five mutants over the five fixes, each checked to fail the suites — including one that puts `shift()` back and is caught by the timeout. No blocking findings outstanding. Nothing fenced. The spec file is untouched.

---

ericeslinger:
**Non-blocking** — `sprout/lang/src/parse.ts:463-464` (`typeExpr`)

`typeExpr()`'s `[` case recurses into itself for the element type but, unlike `primary()`'s `(` case, never re-skips its own bracket when the recursive call fails — `if (element === null) return null;` with no `skipBracketed` call. `primary()` has the equivalent line (`if (inner === null) { this.skipBracketed(')'); return null; }`) specifically to avoid this.

Concretely: `parseRemembers` on `:remembers [a: [[[[[[[[[Ward]]]]]]]]] default oak, b: 3]` (9 `[`, cap 8) returns a non-null `RemembersDeclaration` with `properties: []` and exactly one diagnostic ("Nothing here may be nested more than 8 deep") — `b: 3` is silently dropped and one of the 8 orphaned `]` left behind by the outer `typeExpr` frames is consumed as if it were `:remembers`'s own closing bracket. This is the exact "stray closer" failure mode `skipBracketed`'s own doc comment describes preventing — it just still happens one level of self-recursion away from where the fix reaches.

This is **not a regression from this round**: identical output at `ac1ef6f`, and the same silent drop (via a slightly different path, 2 diagnostics instead of 1) already reproduces at base `main` (`6d799da`), before this PR existed. It's also **not reachable today** through `sprout check`/`parseDeclarations` end-to-end — a message's `carries` type uses the same recursion, but `messageDeclaration`'s `recover()` doesn't care about brackets and silently absorbs the stray closers without losing the next declaration (I checked this case too: `declared: ['b']`, correct). Only a caller that resyncs by looking for one specific closer character — `remembers()`, and later presumably `property()`/object bodies — exposes it, and none of those are wired into real files until B12/B19. Worth a line in a follow-up before then, not a blocker now.

---

**Non-blocking** — `sprout/lang/src/lexer.ts:101-133`, `sprout/lang/src/lexer.spec.ts` (unchanged by this PR)

The cursor rewrite has no new or updated test in its own colocated spec — `lexer.spec.ts` has zero diff in this whole PR. I traced the index arithmetic by hand through partial-consumption, peek-past-buffer, and full-drain-and-reset sequences and found it correct, and couldn't construct a failing case. But per CLAUDE.md, coverage through `parse.spec.ts`'s deep-nesting test doesn't count as this file's own coverage, and that's presently the only thing exercising the drain-and-reset boundary. Given you flagged this file as the one where a defect would be "subtle and everywhere," a direct case in `lexer.spec.ts` — buffer several tokens with `peek(k)`, drain them all via `next()` to force the reset, then `peek`/`next()` again — would pin the boundary down explicitly rather than incidentally.

---

**Non-blocking** — `sprout/lang/src/parse.ts` (`expression`/`unary`/`postfix`/`primary`, unchanged this round)

Found while probing the cap machinery: a host-configured `nesting` cap in the low thousands crashes the compiler with an uncaught `RangeError: Maximum call stack size exceeded` (from deep inside `Parser.expression`/`Lexer.skipBlanks`) instead of a diagnostic — violating this file's own stated invariant ("Problems go to diagnostics; nothing is thrown"). Cause: `expression()`'s precedence climbing recurses ~9 JS frames per nesting level regardless of whether an operator is present, so cap × 9 exceeds V8's default stack well before any performance concern. Reproduced with `caps.nesting: 2999`. This is unchanged since expression parsing was first added earlier in this PR, not something round 3 touched, and needs a cap far beyond anything sane for a limit whose stated purpose is bounding what a person must read — flagging for the record, not asking you to fix it in this PR.

---

**Review**: no blocking findings, 3 non-blocking (pre-existing, not from this round — a `typeExpr` recursion gap that lets a cap-triggered stray closer survive in `:remembers` contexts, not yet reachable end-to-end; `lexer.spec.ts` not updated alongside the cursor rewrite it should directly exercise; a very-large host-configured nesting cap stack-overflows instead of refusing). Gate at `982b2eb`: green, exit 0, 949 passed | 7 skipped (sprout) + 40 passed (cli), matching the PR's receipts exactly. The 52-second regression is confirmed fixed by the lexer cursor specifically — I reproduced 62s/timeout by reverting only `lexer.ts` on top of this round's `parse.ts`, and 237ms at the true head, with no raised timeout anywhere in the repo. `skipBracketed`'s rewrite, all six call sites, and both reworded remedies check out against every attack you asked for.

---

*On the four attacks it was asked to make on `skipBracketed`:*

> **No guard left to get wrong, and no caller worse off:** true for all six call sites. In every one, the caller already consumed the opening bracket before calling `deeper()`/`skipBracketed`, so "consume nothing" never regresses total progress below 1 token — no infinite-loop risk.
>
> **A closer of a different kind inside:** correct, ignored — `skipBracketed`'s depth counter only inspects the two characters it was told about.
>
> **Quadratic or pathological:** no. Every caller advances the lexer monotonically and never rewinds, so any two `skipBracketed` calls scan strictly disjoint stretches; total work is bounded by the token count once, not per nesting level.
>
> **Interaction with `recover()`, `recoverInBraces()`, `afterRefusal`, `capReported`:** none found.

*And on the timing guard, which it verified by reverting rather than by reading:*

> I took `982b2eb`'s `parse.ts` and swapped in `ac1ef6f`'s old `shift()`-based `lexer.ts` — same test: **62074ms, timed out at the default 5000ms**. That's a direct, reproduced before/after: the cursor change is what closes the regression, not a loosened timeout.

*Posted on behalf of the `pr-review` agent, which reviewed at `982b2eb` in a scratch worktree outside the shared checkout. `gh` is not installed in this environment, so it returned its findings rather than posting them itself.*

---

ericeslinger:
## Answering the third review

No blocking findings, and the three non-blocking ones are all pre-existing. One is a house rule I broke and have fixed in `0babc60`; the other two are now issues.

### Non-blocking — `lexer.spec.ts` had no diff

**Fixed.** *"`lexer.ts` changed and `lexer.spec.ts` has zero diff in this whole PR"* is plainly against CLAUDE.md — every source file has a colocated spec that exercises it directly, and coverage through another file's spec does not count. I changed the file I had myself described as the one where a defect would be subtle and everywhere, and then leant on `parse.spec.ts` to cover it.

Six specs now, over what the cursor actually has to get right: the same tokens whichever way they were reached, `peek` and `next` agreeing while interleaved, reading on across the point the buffer empties and resets, reading on when it is drained only part way, and a token's own span and `afterRefusal` surviving a long lookahead.

**And the timing guard had to be sized by measurement, which is the part worth recording.** My first draft peeked fifty thousand tokens and drained them — and a mutant that put `shift()` back **survived it**:

```
shift() drain of  50,000:    222ms     <- inside a 5s timeout; the guard did nothing
shift() drain of 200,000:  3,592ms
shift() drain of 400,000: 18,945ms     <- what the spec uses
```

So the spec uses four hundred thousand and says why, with the count and the timeout both marked not to be touched. It runs in 153ms when the cursor is right. Four mutants over the cursor: the two that matter are caught, and the two that survive are not defects (one is behaviourally identical, one is a memory leak rather than a wrong answer) — which I am reporting rather than counting as coverage.

### Non-blocking — `typeExpr` leaves a stray closer → **#63**

Not fixed here, and an issue instead. The reviewer is right that it is the same failure `skipBracketed` exists to prevent, one level of self-recursion away — and right that it is not a regression: it reproduces on `main` before this PR. The finding's own reachability analysis is what decided it, and I have copied it into the issue: only a caller that resynchronises on one specific closing character is exposed, and the one that is (`remembers()`) is not read from real files until B12 and B19.

**#63** says so, and says the fix and the acceptance test. The reason not to fold it in: this PR has now had three review rounds and two of my own fixes introduced defects, both in exactly this recovery machinery. Adding one more change to the same code, unforced, at the end of that run, is how a seventh round happens.

### Non-blocking — a large host `nesting` cap overflows the stack → **#64**

Also an issue rather than a fix, and this one is mine — it arrived with the expression reader earlier in this PR, not with round three. `parse.ts`'s docblock says *"nothing is thrown"* and three specs assert it, and a host cap in the low thousands makes that false, because precedence climbing costs about nine frames per level whether or not an operator is written there.

**#64** lays out the two honest fixes rather than assuming one: make the expression reader iterative (which `check.ts`'s `typeOf` already is, for this reason), or refuse a cap the parser cannot honour — the second being cheaper but meaning the language tells the host a number, which is the one thing Limits says it never does, so it would need a line in the spec. That choice is not mine to make quietly at the end of a PR.

### On the parts that cleared

Recording these because the reviewer argued them rather than failing to find counterexamples:

- **`skipBracketed` has no guard left to get wrong**, checked at all six call sites, with the no-infinite-loop argument spelled out: every caller has already consumed the opening bracket, so "consume nothing" never regresses progress below one token.
- **It cannot be quadratic**, by a structural argument — every caller advances the lexer monotonically, so any two skips scan disjoint stretches and the total is bounded by the token count once, not per level.
- **The 52-second regression is closed by the cursor specifically**, proved by reverting only `lexer.ts` on top of this round's `parse.ts` and watching the same test take 62s and time out. That is the check I most wanted, because "I made it fast" and "I made the guard stop noticing" look identical from the outside. There is no raised timeout anywhere in the repo, and they grepped for one.

### Receipts

Both at `0babc60`. The `982b2eb` receipts are replaced.

```
npm run gate
```
(exit 0 — 955 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 0babc60)
```

No blocking findings outstanding. Nothing fenced. The spec file is untouched. Merging, with #63 and #64 carrying what is left.

---

PR #65, opened 2026-09-21
B10 and B11: lists, and `let` — Phase 1 closed

### LINKED ISSUE #10: B10 · Lists
**Backlog item B10** · Phase 1 — The type system · size S

## What

Lists: ordered, no duplicates, `add`/`remove`/`includes`/`count`, host-capped, full-list fault

## Spec

Implements: **Properties › Lists** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#6 (B06)

## Blocks

Nothing.

## Working on this

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B10.

### LINKED ISSUE #11: B11 · `let`
**Backlog item B11** · Phase 1 — The type system · size S

## What

`let`: single assignment, no shadowing, allowed wherever an expression is, `let x = spawn …` only where `spawn` is

## Spec

Implements: **Properties › Naming a value** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#9 (B09)

## Blocks

Nothing.

## Working on this

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B11.

ericeslinger:
Closes #10
Closes #11

The last two items of Phase 1. Both small, both mostly discovering that earlier items had already written most of the rule — and then four rounds of review on a third thing that was neither of them.

---

## B10 — lists

**The `listElements` cap has been in `limits.ts` since B02, with the spec's default of 16, and nothing enforced it in either direction.** A written list of any length compiled, and there was no list value for `add` to fault against.

### The value

**Ordered**, in insertion order — *"because `{for … of}` makes it visible in prose"*. A collection that reordered itself would rewrite a sentence an author wrote.

**No duplicates**, so `add` of something held and `remove` of something absent each do nothing. Neither is an error: a list is a statement about what is true, and saying a true thing twice is not a mistake. They hand back **the same list**, so a caller can tell nothing happened.

**Bounded**, and adding a *new* element to a full one **faults**:

> `adjust` clamps at a ceiling because reaching the ceiling is the meaning; dropping an element would lose something the author wrote.

`ListFull` is thrown, the same shape as `BudgetExhausted`, for B34 to turn into the world's `fault` passage.

### The cap, on both sides of the compile/run line

A written literal past it is **refused** — said once, at the element that breaks it, then read on, so an author whose list is four too long is still told about the capital letter in it. Refused rather than truncated: a silent drop is the one thing a full list must never be.

### One row of the checker's table widened

*Lists* names four operations including `count`; the checker's table names `count` for a container or a set role only. The fuller sentence wins. `count(K)` stays refused on a list, which holds values and has no contents composing a kind.

---

## B11 — `let`

**Its type is the expression's exactly.** An integer 0 to 9 stays an integer 0 to 9; a name for a thing narrowed by `is()` stays at the kind it was narrowed to.

**Written once and never again** — and B11 adds no rule of its own. There is no reassignment anywhere in the language, so the only way to write a name twice is to `let` it twice, which `Scope.introduce` already refuses as shadowing. The review confirmed by `git grep` that `checkLet` is the only production caller of `introduce`.

**It cannot name something that changes the world**, because an initializer is an expression — also refused by machinery that already existed. A refused `let` does not come into scope.

**No type may be written.** *"Its type is the expression's, exactly, so nothing is annotated"* says there is nothing to write, not that writing one is refused; `let n: integer = 1` is refused, which keeps a second spelling of a type out of the language.

---

## The part that took four rounds

Enforcing the list cap broke a `min`/`max` bound: reading a bound meant reading *any* literal and complaining about its shape afterwards, so whatever went wrong inside answered for the bound. `min [… seventeen things …]` said *"A list holds at most 16 things"*, whose remedy is no use to someone who should not have written a list there.

**Three attempts to fix that each broke an input beside the one they fixed.**

| attempt | what it lost |
| --- | --- |
| peek, refuse without consuming | the abandoned construct's closer stayed behind; `:remembers [visits: 0 min [1, 2], walks: 1]` lost `walks` |
| refuse, then step over what was written | stepping over a bare `]` ate the `:remembers`'s **real** terminator; `[handled: false, visits: 0 min ]` lost `handled` — worse than base |
| stop at any closer | a **decoy** closer survives; `[handled: false, visits: 0 min ], walks: 1]` lost `walks` again |

The last two are the two horns of one ambiguity, and the fourth review named it:

> a reader that has not consumed anything cannot tell a stray closing bracket from the one its own caller is waiting for.

That is missing information, not a bug in the attempts. Resolving it needs the enclosing closer threaded down to the point of refusal — three signatures and every caller — and making a design change at the end of a four-round review is how a fifth round happens.

**So the bound reader is reverted to what `main` does.** Noisier, and it loses nothing that was not already lost. **#67** carries both halves of the cost, with all three failed attempts written down.

Worth stating plainly: the thing being fixed across those rounds was a **non-blocking message-quality finding**, on a path no real world can reach until B12/B19 wire `property()` into `file()`. It cost three regressions, one strictly worse than base. I should have stopped after the first.

## The net from #66 lands here

Against what I wrote in #66, because it is **test-only and cannot cause an eleventh regression**. One invariant, over all three readers:

> given a well-formed item, a defective item, and a well-formed item, every well-formed item is either **kept** or **named** in something said.

It needs no knowledge of what the defect was — the spec builds the shapes. That is the property all ten bugs violated and the one none of the three existing nets could express: they assert no-throw, no-repeat and determinism, and **a parser that drops an entry while saying one true thing passes all three.**

Reconstructions of the #58 and #62 failures were run against it; both caught. It also caught two errors in itself first — `[oak, enum]` is a legal list, and a bare `]` ends the construct rather than sitting inside it — the first of those found by its own assertion that every shape must really be a defect.

The two shapes it excludes are **named in the spec**, not quietly omitted, with a pointer to #67.

## Notes recorded

For B10: `count` on a list is in one place and not the other; the `listElements` cap is a refusal at compile and a fault at run, and the limits table describes only the first; what a full list does when added to with something it already holds; and **how two lists are compared**, which the spec never says — same element type, same things, same *order*, because order is visible in prose and comparing them as sets would make `==` disagree with what an author can see.

For B11: whether a `let` may be annotated (no) and what it may be called (a lower-case word). Three parts have nowhere to land — *where* a `let` may be written (B24, B29), `let x = spawn …` (B18), and what it costs in steps (B34) — recorded so those items know they inherit them.

## Receipts

Both at `61fda00`, the PR head.

```
npm run gate
```
(exit 0 — 1003 passed | 7 skipped across lang/core/stores, 40 passed in cli; 955 before this PR)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 61fda00)
```

Nothing fenced. The spec file is untouched. **#66** and **#67** carry what is left.


ericeslinger:
**Blocking** — `sprout/lang/src/lists.ts:139`

`SproutList.equals` is a fifth operation the spec doesn't name, and its scoping (order-sensitive) isn't recorded as a hole.

The Lists section (`docs/design/sprout-design-spec.md:521`) says "Four operations: `includes(x)`, `count`, `add` and `remove`," and this file's own doc comment repeats "Four operations, and no more" — `equals` isn't one of them, and nothing in the spec says how two lists should compare (order-sensitive structural equality, as implemented, vs. order-insensitive, vs. reference identity). The only spec text that speaks to value-type comparison at all is line 1358, under "What an extension may add" — about host-defined library types, not the language's own built-in list. `equals` is deliberately built (two dedicated tests in `lists.spec.ts`, called out as mutation-tested in the commit message) but the working notes' "Found while building lists (B10)" section records three other holes for this same item and says nothing about this one.

Add a line under that heading recording that list-equality semantics aren't stated by the spec and are being read in from the general "a value type must declare how it compares" principle — or say in the PR body why that's not needed.

---

**Non-blocking** — `sprout/lang/src/parse.ts:728-730`

The list cap changes what a `min`/`max` bound reports when the bound is a list literal that's also over cap, and the new message's remedy doesn't fix the actual problem.

Repro: `:x integer default 0 min [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17]` (17 elements, default cap 16). At base `2c31f93` this reports `A min is a whole number.` At head `90d8511` it instead reports `A list holds at most 16 things.` with remedy `Take some out, or hold them somewhere that is not a list.` — following that remedy down to 16 items still leaves a list where an integer is wanted; the real message only appears once the list is *also* short enough to parse. `max` is symmetric. This is because `propertyBody`'s `if (bound === null) return null;` (line 729) now fires before `bound.kind !== 'integer'` ever runs, whenever the mistyped bound is also over cap. No spec in `parse.spec.ts` covers a list literal written as a min/max bound at all, so nothing caught the change.

---

**Non-blocking** — `sprout/lang/src/parse.ts:683-684`, `sprout/lang/src/parse.spec.ts:1096-1131`

No test checks the span of the new list-cap diagnostic, only its text.

Verified by mutation: changing the reported span from `element.at` to `open.at` (reporting at the list's opening bracket instead of the offending element) still passes the full suite (1005/1005, including the four `listElements` tests). A future edit that moves the report to the wrong token would ship undetected. Pin at least one existing cap test to the element's location.

---

**Non-blocking** — `sprout/lang/src/parse.ts:639-690`

Two over-cap list literals in the same declaration aren't tested, and nothing guards the scoping of `overCap` as a per-call local rather than a per-declaration field.

Verified by mutation: turning `overCap` into a parser field reset the same (never, within one `parseRemembers` call) as the adjacent, pre-existing `capReported` nesting flag — a natural mistake to copy, given it sits two lines above and looks like the same pattern — still passes the full suite. Under that mutation, `:remembers [a: [1,2,3,4], b: [5,6,7,8]]` (cap 3) reports the cap once for `a` and silently drops `b` with no diagnostic — the silent drop this PR's own comments say a full list must never produce. Add a test with two list-valued entries in one `:remembers`, both over cap, asserting two independent reports.

---

**Review**: 1 blocking, 3 non-blocking. Gate at `90d8511`: green (993 passed | 7 skipped sprout, 40 cli, corpus as expected — matches the PR's receipts). List-cap and `let` logic held up under ~55,000 fuzzed/property-based trials against the built parser and checker — no live behavioral defect found there, and the four list operations and their order-preservation match the spec. The findings are a spec-silent `equals` left off the working notes, one real adjacent-code-path diagnostic regression (min/max bounds given an over-cap list), and two suite gaps surfaced by targeted mutation rather than by reading.

---

*On the three questions it was asked, none of which became findings:*

> **`ListFull` thrown from `SproutList.of`, not just `add`**: defensible. Nothing in production code calls `.of()` yet — it's B16's future entry point — and throwing keeps one failure mode with `add`, which is right for what's explicitly framed as an invariant violation rather than a normal outcome.
>
> **`let` and `Scope.introduce`**: confirmed by `git grep` — `checkLet` is the only production call site of `Scope.introduce` at this PR's head (the only other scope mutator, `narrowing()`, is explicitly the one deliberate non-shadowing re-bind). The claim "B11 adds no rule of its own; it discovers B08 already wrote it" checks out. 3000 randomized `let` trials found no crash, no silent refusal, and no scope corruption on a refused `let`.
>
> **`let n: integer = 1` and `let Ward = 1`**: both defensible readings, and — unlike the `equals` case — both properly recorded in the working notes with reasoning.

*And on its own method, which is worth keeping:*

> Hand-tracing the list-cap loop turned up several *apparent* bugs that I could not reproduce and eventually disproved by running the real parser: every one turned out to still carry a diagnostic, just not always at the level I first expected.

*Posted on behalf of the `pr-review` agent, which reviewed at `90d8511` in an isolated worktree outside the shared checkout. `gh` is not installed in this environment, so it returned its findings rather than posting them itself.*

---

ericeslinger:
## Answering the review

All four taken, in `fec11c2`. Each reproduced before being touched.

### Blocking — `equals` is a fifth operation, unrecorded

**Taken, and the framing is the one that matters.** I wrote *"Four operations, and no more"* in the file's own doc comment and then built a fifth one twelve lines below it. `equals` is genuinely needed — `a == b` on two lists of one type is legal by the checker's own table, and something has to answer it at run time — but **"needed" is not "stated"**, and this repository's whole discipline is that the difference gets written down.

The notes carry it now: the same element type, the same things, in the same **order** — because order is insertion order and `{for … of}` makes it visible in prose, so two lists that render differently are not the same list. The alternative, comparing them as sets, would make `==` disagree with what an author can see on the page.

I also said what it is, since the doc comment's "four and no more" is otherwise a lie: `equals` is not an operation offered to an author, it is what the runtime uses to answer `==`.

### Non-blocking — the min/max regression

**This is the seventh time a change to this parser has broken the input adjacent to the one it changed**, and the second time in this session that my own fix did it. Confirmed exactly as reported:

```
:x integer default 0 min [… seventeen things …]
  before   A min is a whole number.
  after    A list holds at most 16 things.   (remedy: take some out)
```

Taking some out still leaves a list where a number goes.

The cause is worth naming, because it is not really the cap: `propertyBody` read a bound by **reading any literal and complaining about its shape afterwards**, so whatever went wrong *inside* the literal answered for the bound. The cap just gave that a new way to go wrong.

So a bound now asks what is written before reading it. That fixes the regression and improves what was always there — `min oak`, `min "9"` and `min true` are refused at the word rather than after building a literal nobody wanted — while `min 1.5` keeps `literal()`'s own better sentence (*"Sprout has no fractions"*) rather than being called not-a-number, and a negative bound still reads. Five specs, one of them specifically that the fraction message survives.

### Non-blocking — the span was never pinned

**Taken.** The reviewer's mutant is right: moving the report from the offending element to the list's opening bracket passed all 1005 tests. A test now asserts the span covers `e16`, and that mutant fails it.

### Non-blocking — `overCap` scoping was never guarded

**Taken**, and this is the sharper of the two, because the mistake it guards against is one a reader is *invited* to make: `overCap` is a per-call local sitting two lines below `capReported`, which is a per-declaration field, and they look like the same pattern. Copying it reports the first over-cap list and drops the second in silence — the one thing this PR's own comments say a full list must never do.

There is a test now with two over-cap lists in one `:remembers`, and the real mutant fails it.

**One correction to my own work here:** my first attempt at reproducing that mutant was `let overCap = this.capReported && false`, which is just `false` — a no-op that changed nothing and survived correctly. I rewrote it as an actual per-parser field before trusting the result. Reporting it because a mutant that survives for being a no-op looks exactly like a gap in the suite, and I have now made that mistake twice.

### On what cleared

- ~55,000 fuzzed and property-based trials against the built parser and checker, including a property that a returned list-literal node never exceeds the cap. No behavioural defect in the cap loop itself.
- 3,000 randomised `let` trials, and a `git grep` confirming `checkLet` is the only production caller of `Scope.introduce` — so "B11 adds no rule of its own, it discovers B08 already wrote it" is checked rather than asserted.
- `ListFull` thrown from `SproutList.of` judged defensible: nothing calls it yet, it is B16's entry point, and throwing keeps one failure mode with `add`.

And the method note, which I want on the record because it is the opposite of this file's usual story:

> Hand-tracing the list-cap loop turned up several *apparent* bugs that I could not reproduce and eventually disproved by running the real parser.

Reading produced false positives; running produced the true ones. Both directions of that are worth knowing.

### Receipts

Both at `fec11c2`. The `90d8511` receipts are replaced.

```
npm run gate
```
(exit 0 — 998 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at fec11c2)
```

No blocking findings outstanding. Nothing fenced. The spec file is untouched.

---

ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:760` (`integerBound`)

Refusing a bracket-shaped bound without consuming it reintroduces #62's stray-closer bug in a new place: the abandoned list's own `]` gets mistaken by `remembers()`'s loop for the `:remembers`' own closer, ending the parse early and silently dropping every entry after it.

`parseRemembers(':remembers [visits: 0 min [1, 2], walks: 1]')` — at base `2c31f93`: 1 refusal ("A min is a whole number."), `properties: [walks]`. At this head: 4 refusals ("A min is a whole number.", "A remembered property starts with its name, and `[` is not one.", "...the number 1 is not one.", "...the number 2 is not one."), `properties: []`. Both `visits` and `walks` vanish, and nothing says `walks` went missing — it was well-formed. Any list-shaped bound reproduces it (`[1]`, `[1,2]`, `[1,2,3]` all do it), not only an over-cap one. The same leftover-token pattern also adds a false second diagnostic for a bareword bound: `:remembers [visits: 0 min oak, walks: 1]` says "`oak` needs a colon between its name and its value." in addition to the real complaint, where base said only the one true thing (`walks` is still recovered in that narrower case, just with extra noise).

This is exactly the shape `describe('#60 — the shape every round of bugs here has turned on, enumerated')` and the POOL fuzzer beside it were built to catch, but both only exercise a property/remembers-entry's *value*, never its `min`/`max` bound — so this instance passes both, and swallowed-entry bugs wouldn't trip the fuzzer's invariants (no-throw, no-duplicate-diagnostic) even if it did generate the shape.

Fix shape: when the peeked token opens a bracketed construct, step over it (`skipBracketed`, as `literal()`'s own `[` branch already does) before refusing, instead of leaving it for the enclosing loop to walk into token-by-token.

---

**Non-blocking** — `sprout/lang/src/parse.ts:760` (`integerBound`)

For a bound written as a bare `-` with nothing after it (`min -`) or a doubled sign (`min --3`), the peek-first check never reaches `literal()`'s more specific message. Base `2c31f93` said "A minus sign needs a number after it." (remedy "Write a whole number, as in `-3`."); this head says the generic "A min is a whole number." (remedy "Write `min 0`, or leave it out."). Both are true and both have a remedy, but the new one no longer tells an author who was mid-way through a negative number that the digits are what's missing. Not covered by the new suite in either direction.

---

**Review**: 1 blocking, 1 non-blocking. Gate at fec11c2: green (exit 0; 998 passed | 7 skipped in sprout, 40 in cli — reproduced independently in a fresh worktree). The `equals`/list-comparison working-notes line, the cap diagnostic's span assertion, and the two-over-cap-lists test all check out: `equals()` matches the note's own words and `identityType`/`sameType` do make two same-typed lists comparable as the note assumes, and I broke each of the two parse.ts properties those tests guard (moved the cap's span to the list's open bracket; made the over-cap flag a persistent field instead of a per-list local) and watched the matching new test fail each time. The `min`/`max` fix itself over-corrects: peeking then refusing without consuming reintroduces, against a list-shaped bound inside `:remembers`, the exact stray-token recovery bug this file has hit four documented times before.

---

*Also checked and found nothing wrong with:* `min` at end of file, `min` before `max`, `+3`, `0x10` (identical on both sides of the diff — pre-existing, not from this change), a duplicate `min`/`max`, a bound on a typeless or non-integer property, and `min 1.5` / `max 1.5` / `min -3.5`, all of which correctly defer to `literal()`'s fraction message.

*Posted on behalf of the `pr-review` agent, which re-reviewed at `fec11c2` in worktrees outside the shared checkout, all removed afterwards. `gh` is not installed in this environment, so it returned its findings rather than posting them itself.*

---

ericeslinger:
## Answering the re-review

Both taken, in `97cdd20`. This is the round I asked for, and it found exactly what I asked it to look for.

### Blocking — my fix for the last round's finding broke the input beside it

**The eighth time a change to this parser has broken an adjacent input, and the second round running in which the breakage arrived in my own fix for the round before.** Confirmed exactly as reported:

```
:remembers [visits: 0 min [1, 2], walks: 1]
  before   kept nothing, 4 diagnostics, and `walks` — written correctly — vanished
  after    kept walks, one diagnostic: A min is a whole number.
```

**Fixed as suggested**, and the diagnosis is the part worth keeping: *"the same stray-token recovery bug this file has hit four documented times before."* `skipBracketed` was written two rounds ago for precisely this, with a doc comment explaining precisely this, and I did not carry it across to the new refusal path. Knowing the lesson and writing it down was not enough to apply it.

A refused bound now steps over the one thing that was written, and steps over a **bracket whole**. Every bad bound shape — `[1, 2]`, `[[1]]`, `oak`, `"9"`, `true`, `Ward`, `:wet`, `{`, `}` — keeps the entry after it and says one true thing. That is better than the code *before* the round that introduced this, where a bareword bound kept the entry but added a false second complaint about it.

**And the note about why the existing nets missed it is the most useful sentence in the review:**

> `describe('#60 — the shape every round of bugs here has turned on, enumerated')` and the POOL fuzzer beside it … only exercise a property/remembers-entry's *value*, never its `min`/`max` bound — so this instance passes both, and swallowed-entry bugs wouldn't trip the fuzzer's invariants (no-throw, no-duplicate-diagnostic) even if it did generate the shape.

That is two separate gaps: the input space never reaches bounds, and **no invariant in that suite is about an entry going missing** — which is the failure mode this file keeps producing. The specs added here cover the bound path directly; the second gap is bigger than this PR and I have not closed it.

### Non-blocking — a minus sign lost its own sentence

**Taken.** `min -` and `min --3` are someone part-way through writing a negative number, and *"A min is a whole number"* does not tell them the digits are what is missing. A minus sign goes back to `literal()`, which owns that sentence:

```
:x integer default 0 min -     A minus sign needs a number after it.  (Write a whole number, as in `-3`.)
```

### One shape that looked like a regression and is not

While checking, I found that an unclosed `:remembers` whose last entry failed no longer says it is unclosed — `:remembers [a: 0 min oak` says only *"A min is a whole number."*

I checked before assuming it was mine, and it is not: that is the pre-existing rule that a failed entry read at the end of a file has already explained itself. `:remembers [a: 1.5`, which touches no bound at all, behaves identically, while `:remembers [Zeta` says both because its failure does not reach the end. My change only added one more way for an entry read to fail. Leaving it as it is, because changing it means revisiting the #60 rule about not repeating "never closed" once per level, and that is not this PR's to reopen.

### On the confirmations

The reviewer re-ran my own two mutants rather than taking my word — moving the cap's span to the opening bracket, and making the over-cap flag a persistent field — and watched each new test fail. That is the check I most wanted, since a test that passes both ways is the failure mode I have hit twice in this session.

### Receipts

Both at `97cdd20`. The `fec11c2` receipts are replaced.

```
npm run gate
```
(exit 0 — 1001 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 97cdd20)
```

Four mutants over the two fixes, each checked to fail the suites. No blocking findings outstanding. Nothing fenced. The spec file is untouched.

---

ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:794-798` (`skipWritten`), surfacing at `sprout/lang/src/parse.ts:855`

`skipWritten` consumes the offending token unconditionally, including when that token is a bare closer that is not stray at all but is the enclosing list's own `]` — and when that happens, `remembers()`'s `if (this.done) return null` at line 855 discards every property already collected, with nothing said about the loss.

Repro: `parseRemembers(new SourceFile('k.sprout', ':remembers [handled: false, visits: 0 min ]'), diagnostics)` returns `null`, and `diagnostics.refusals` contains only `"A min is a whole number."` — `handled`, a well-formed property already parsed and ready to keep, is dropped with no mention of it or of the list being cut short. I checked the same source at `fec11c2` (this PR's previous head, before `skipWritten` existed) and at base `main` (`2c31f93`): both correctly return `{ properties: ['handled'] }`. This is a regression introduced by this round's diff, not present before it, and it contradicts the claim that the new behaviour is better than base for every shape. It is also exactly the defect class `skipWritten`'s own docstring says it exists to prevent ("dropping an entry that was written correctly, with nothing said about it") — reintroduced through the one case it can't tell apart from ordinary junk: a lone closer that the construct around it actually needed. The comment justifying line 855 ("a file which ran out inside the entry has already been explained by whatever read it") is the assumption that breaks: the file didn't run out; `skipWritten` manufactured that appearance by eating the real `]`.

A `skipWritten` that declines to consume a bare closer — leaving it for the enclosing `]`/`)`/`}` check to find, the way the pre-fix code did by accident — would restore this.

---

**Blocking** — `sprout/lang/src/parse.ts:770-773` (`integerBound`, the minus/digit branch)

Anything beginning with `-` is now handed unconditionally to `literal()`, which never calls `skipWritten`/`skipBracketed` — so a `-` bound followed by a bracket is only partly stepped over (the `-` alone), and the enclosing loop's own one-token-at-a-time recovery walks through the leftover bracket and mistakes its closer for the list's own, dropping every entry after it.

Repro: `parseRemembers(new SourceFile('k.sprout', ':remembers [visits: 0 min -[1], walks: 1]'), diagnostics)` returns `{ properties: [] }` — `walks`, written correctly, never appears, and none of the three refusals produced ("A minus sign needs a number after it.", "...`[` is not one.", "...the number 1 is not one.") says so. Same for `-[1, 2]` and `-[[1]]`. This specific outcome is not a new regression — the identical source loses `walks` at `fec11c2` and at base `2c31f93` too, via the other (pre-`skipWritten`) branch — but this round's diff is exactly what widened the minus branch's condition (dropping the old `this.peek(1).kind === 'integer'` lookahead) to route every `-`-prefixed bound through `literal()` instead of through the branch `skipWritten` now protects, and the new specs never test a `-` bound against a bracket, or against a following sibling, so the hole is completely unguarded by this PR's own tests. Non-bracket minus mistakes are all fine: `-oak`, `-"9"`, `-true`, `-Ward`, `- - 3`, and a trailing `-` each leave the sibling intact with exactly one thing said — it's specifically `-` followed by `[` that breaks.

Routing the minus branch's failure through `skipWritten` (or having `literal()`'s minus-branch fall through to it) would close this the same way the `else` branch was closed.

---

**Non-blocking** — `sprout/lang/src/parse.ts:794-798` (`skipWritten`)

`skipWritten` steps over exactly one bare token, or one whole bracket if that token opened one — but a bad bound written as two or more bare tokens in a row still produces more than one diagnostic, undercutting the "says one true thing about it, not one per token it stepped over" claim the new spec makes for this code.

`:remembers [visits: 0 min max 9, walks: 1]` says both `"A min is a whole number."` and `"A remembered property starts with its name, and the number 9 is not one."`; `:remembers [visits: 0 min self.count, walks: 1]` says three things. `walks` survives in every case tried — this is not a data-loss bug, just noise: the second (and third) message is a red herring about leftover junk rather than about the bound. Not covered by the new bad-bound array, which is entirely single bracketed-constructs or single bare tokens.

---

**Review**: 2 blocking, 1 non-blocking. Gate at `97cdd20`: green, exit 0 (1001 passed | 7 skipped in sprout, 40 in cli — matches the PR's receipt exactly). `git diff fec11c2..97cdd20` is the claimed small change (`integerBound` defers to `literal()` for a digit or minus, else refuses and calls new `skipWritten`/`CLOSER_OF`), and it fixes the round-2 finding for every bracketed, non-minus shape — but a bare closer written as a bound can itself be the enclosing list's real close (dropping everything collected so far, comment 1), and the minus branch bypasses the new machinery entirely when a bracket follows (comment 2), so the "entry after it survives" claim doesn't hold for those two shapes. Verified the 3 named mutants of `skipWritten`/`CLOSER_OF` (consume nothing, consume one token without the bracket step, invert `CLOSER_OF`) plus a 4th (invert the guard) — all four killed by the new specs, independently confirmed. Did not re-raise the structural gap tracked by #66.

---

*Also confirmed, and not raised as findings:* an unclosed `[` as a bound still loses the sibling, but identically at base — not a regression; `[message foo]`, `[enum Ward {oak}]` and `[walks: 1]` are all stepped over whole correctly; `skipBracketed`'s "take nothing when unclosed" is harmless on its own; and `min (`/`min (1` recover fine through the loop's own comma-eating.

*Posted on behalf of the `pr-review` agent, which reviewed at `97cdd20` in throwaway worktrees at three commits, all removed afterwards. `gh` is not installed in this environment, so it returned its findings rather than posting them itself.*

---

ericeslinger:
## Answering the third review

All three taken, in `5484a7e`. **This is the ninth adjacent-input break on this parser, and the third round running in which it arrived in my own fix for the round before.**

### The two blocking findings are two halves of one mistake

And each half was learnt by breaking the other, which is worth setting out plainly because it is the whole story of this function:

```
LEAVING a closer behind   →  the enclosing reader takes it for its own and ends early
  :remembers [visits: 0 min [1, 2], walks: 1]      lost walks        (round two)

EATING a closer           →  the enclosing construct looks as though it ran out,
                             and everything already read is thrown away
  :remembers [handled: false, visits: 0 min ]      lost handled      (round three)
```

The second loses more, and the reviewer's diagnosis of *why* is exact: *"the comment justifying line 855 — 'a file which ran out inside the entry has already been explained by whatever read it' — is the assumption that breaks: the file didn't run out; `skipWritten` manufactured that appearance by eating the real `]`."* `handled` was written correctly and has nothing to do with a bound two entries later.

**So `skipWritten` is one rule now**, rather than a list of cases: take everything up to a comma, any closing bracket, or the end of the file, stepping over an *opening* bracket whole. The two failures are the two ends of that rule, and stating it once is what makes them both impossible rather than each patched.

That also answers the non-blocking finding for free: between those stopping points it takes all of it, so `min max 9` and `min self.count` are one complaint rather than one per leftover word.

### And the minus branch reaches it

The reviewer is right about the provenance, and it is the more uncomfortable version: `min -[1]` was as broken before this PR as during it, but **this round's change is what widened the minus branch** — dropping the `peek(1).kind === 'integer'` lookahead — so every `-`-prefixed bound now routes through `literal()`, which owns the sentence about minus signs but does not step over what follows one. Widening a path without carrying its protection across is the same mistake as round two, one function along.

It calls `skipWritten` now.

### Where this leaves it

Twenty-six bad bound shapes — brackets, nested brackets, a bracket holding a declaration keyword, barewords, kinds, symbols, strings, bare closers, several words, every minus form, fractions — each keeps the entry written after it and says **exactly one true thing**. The two that still drop are an unclosed bracket the author wrote, which behaves as it does on `main`.

Six mutants over the fix. Five fail the suites; the sixth **hangs**, which is its own kind of caught — consuming past the end of the file never terminates, since `next()` returns `end` for ever. I am reporting that rather than counting it as a clean kill.

### On the review itself

Three things it did that I want on the record, because they are why this found anything:

- It checked the bare-closer repro against **both** `fec11c2` and base `2c31f93` before calling it a regression, rather than assuming.
- It separated "this is broken" from "this PR broke it" for the minus case, and said so — which changed what I had to fix but not whether.
- It re-ran my own four mutants independently instead of taking the claim, and noted that **none of them touches the minus-sign spec**, which is consistent with that spec never exercising `skipWritten` at all. That is the observation that made the second finding findable.

### Receipts

Both at `5484a7e`. The `97cdd20` receipts are replaced.

```
npm run gate
```
(exit 0 — 1002 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 5484a7e)
```

No blocking findings outstanding. Nothing fenced. The spec file is untouched. #66 carries the structural gap that would catch this class rather than its instances.

---

ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:811`

`skipWritten`'s stop check runs before it has consumed anything, so when a bad bound is immediately followed by a closing bracket, the loop cannot tell "this closer is the enclosing `:remembers`'s own terminator" from "this closer is part of what was written badly, and the real terminator is still ahead." It leaves the closer either way. The enclosing loop gets the first case right and the second case wrong — it mistakes the decoy for its own close and silently drops everything after it.

Take the PR's own "does not eat a closing bracket" fixture and add one property after the bad one: `:remembers [handled: false, visits: 0 min ], walks: 1]` gives `properties: ['handled']` and only `"A min is a whole number."` — `walks` vanishes with nothing said about it. At `97cdd20` the same input kept both (`['handled', 'walks']`). Not limited to an empty bound: `min oak]` (junk, then a decoy `]`) drops the same way. The same up-front check also regresses the "one true thing" claim when the bad bound is itself a stray `}` or `)`: `min }` said one thing at `97cdd20` and says two at this head, because the enclosing loop's safety net has to force a token through on a second pass before it recovers. This is the exact failure this PR's docstring names — "the loop reading around this takes it for its own and ends early" — reintroduced by the rule written to end it.

---

**Review**: 1 blocking. Gate at `5484a7e`: green — reproduced in a clean worktree with `npm ci && npm run gate`, exit 0, 1002 passed | 7 skipped (sprout), 40 passed (cli), matching the PR's receipt exactly. The collapse to one rule is real progress — verified equal-or-better against `97cdd20` across ~35 bad-bound shapes, including two it outright fixes (`-[1]`, `-[1, 2]` used to drop the next entry, now don't) — but a bound immediately followed by a decoy closing bracket still drops the entry after it, or reports two things instead of one, exactly the failure class this round exists to close.

---

*Its own notes on what it checked and cleared, which are the reason the finding is trustworthy:*

> **Termination:** clean. Every loop iteration does exactly one unconditional `this.next()` after the two early-return checks, and the lexer's `end` token is documented and confirmed sticky.
>
> **Minus branch:** clean, checked by execution. `min 1.5`, `min -1.5`, a fraction directly before a comma, and a fraction at EOF all produce exactly one diagnostic and leave the cursor exactly at the following comma/EOF. A genuine improvement on messier shapes — `min 9.0.0` went from 3 diagnostics to 1.
>
> **Mismatched bracket types** (`min (1, 2]`) also lose the later entry, but I ran that exact input against `97cdd20` and got the same drop — pre-existing, and squarely the pattern #66 already names, so I did not turn it into a separate finding.

*And on why it raised this at all, given #66 exists:*

> I'm treating "the structural gap tracked by #66" as the *absence of systematic invariant testing*, not as blanket cover for any instance of the pattern it was opened to prevent — a 9th concrete instance, freshly introduced by this round's own fix, seemed exactly what you were asking me to surface rather than wave through.

*Posted on behalf of the `pr-review` agent, which reviewed at `5484a7e` in worktrees outside the shared checkout, all removed afterwards. `gh` is not installed in this environment, so it returned its findings rather than posting them itself.*

---

ericeslinger:
## Answering the fourth review — and stopping

The finding is real and I have reproduced it. **It is the tenth adjacent-input break on this parser, and the fourth round running in which it arrived in my own fix for the round before.**

**I am not fixing it again.** `61fda00` reverts the bound reader instead.

### Why reverting is the right answer and patching is not

The last two rounds are the two horns of one ambiguity, and the review named it better than I did:

```
LEAVING a closer behind   →  the enclosing reader takes it for its own and ends early
                             lost `walks`                                  (round 2)
EATING a closer           →  the enclosing construct looks as though it ran out
                             lost `handled`                                (round 3)
STOPPING at any closer    →  a decoy closer survives and the real one is still ahead
                             lost `walks` again                            (round 4)
```

A reader that has not consumed anything **cannot tell a stray closing bracket from the one its own caller is waiting for.** That is not a bug in my three attempts; it is missing information. Resolving it means threading the enclosing closer down to the point of refusal — three signatures and every caller — and making a design change at the end of a four-round review is precisely how a fifth round happens.

So the bound reader goes back to what `main` does: read a value, complain about its shape. Noisier, and **it loses nothing that was not already lost.**

**#67** carries both halves of what that costs — a bound reporting what is wrong *inside* it, and `min -[1]` losing the entry after it — with all three failed attempts written down, so the next person does not rediscover them in the same order I did.

Worth being plain about the accounting: the thing I was fixing across three rounds was a **non-blocking message-quality finding**, on a code path no real world can reach yet, since `property()` is not wired into `file()` until B12/B19. It cost three regressions, one of them strictly worse than base. That is a bad trade and I should have stopped after the first one.

### The net from #66 lands here instead

I said in #66 that folding it into this PR would be wrong. I have changed my mind, and the reason is narrow: **it is test-only, so it cannot cause an eleventh.**

One invariant, over all three readers:

> given a well-formed item, a defective item, and a well-formed item, every well-formed item is either **kept** or **named** in something said.

It needs no knowledge of what the defect was — the spec builds the shapes, so what was well formed is known. That is the property all ten bugs violated, and the one none of the three existing nets could express: they assert no-throw, no-repeat and determinism, and **a parser that drops an entry while saying one true thing passes all three.**

I reconstructed the #58 and #62 failures and ran them against it. Both caught.

And it caught two errors in *itself* before it caught anything else: `[oak, enum]` is a legal list, because nothing reserves an option's name; and a bare `]` ends the construct rather than sitting inside it, so what follows is not a sibling. The assertion that every shape really *is* a defect is what found the first — the second time this suite has needed telling that, and the reason it is in there.

The two shapes it excludes are **named in the spec**, not quietly left out, with a pointer to #67.

### On the review

Three things it did that earned the finding:

- It attacked the rule by **reasoning about what the rule cannot know**, then confirmed by execution — its own note says the sweep of ~35 shapes missed this and the targeted attack found it.
- It distinguished "head is worse than base" from "this was always broken" for the mismatched-bracket case, and declined to raise the second.
- It pushed back on my own instruction. I told it #66 covered the structural gap; it answered that #66 covers *the absence of a net*, not any instance of the class, and that a fresh instance introduced by this round was exactly what I had asked for. That is right, and I would have shipped the bug if it had taken me at my word.

### Receipts

Both at `61fda00`. The `5484a7e` receipts are replaced.

```
npm run gate
```
(exit 0 — 1003 passed | 7 skipped across lang/core/stores, 40 passed in cli)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 61fda00)
```

No blocking findings outstanding. Nothing fenced. The spec file is untouched. #66 and #67 carry what is left, and #66's first cut is in this diff.

---

Issue #66, opened 2026-09-22
The parser's nets have no invariant about a well-formed item going missing


ericeslinger:
Not a backlog item. It comes out of the re-review of #65, and out of counting.

## The pattern

**Eight times** now, across #58, #59/#60, #62 and #65, a change to `parse.ts` has broken the input *beside* the one it changed. Twice in #62 and twice in #65 the breakage arrived in my own fix for the round before. Every one of them was found by a reviewer running the built parser, never by the suite.

The single most common shape is the worst one the language has: **a well-formed item vanishes with nothing said about it.** `:remembers [visits: 0 min [1, 2], walks: 1]` losing `walks`; `:x [oak, Zeta silver]` losing `silver`; `:remembers [a: <nine deep>, b: 3]` losing `b`.

## Why the existing nets miss it

Two of them exist and both are blind to this, for different reasons. From the review that found the latest instance:

> `describe('#60 — the shape every round of bugs here has turned on, enumerated')` and the POOL fuzzer beside it only exercise a property/remembers-entry's *value*, never its `min`/`max` bound — so this instance passes both, and swallowed-entry bugs wouldn't trip the fuzzer's invariants (no-throw, no-duplicate-diagnostic) even if it did generate the shape.

Two separate gaps:

1. **The input space is too narrow.** Both nets generate values, and a declaration has more parts than its value — a written type, a `min`, a `max`, a name, a separator. The enumeration was built for the nested-bracket shape and never widened.
2. **No invariant is about the failure mode.** The generator asserts three things: it never throws, it never says one thing twice in one place, and the same source gives the same answer twice. **None of them is violated by dropping an entry in silence.** A parser that returns `[]` and says one true thing passes all three.

## What to build

An invariant, checked over generated input, of roughly this shape:

> Given a declaration built as *(a well-formed item, an item with one deliberate defect, a well-formed item)*, every well-formed item is either **kept** or **named in a diagnostic**. Nothing may disappear unexplained.

That is the property every one of the eight bugs violated, and it is checkable without knowing what the defect was: the generator builds the items, so it knows which ones were well-formed.

It needs the input space widened alongside it — the defect should be injectable into any part of a declaration, not only into a value: a type, a `min`/`max` bound, a name, a separator, an option, a nested list at any depth.

Worth doing over `parseDeclarations`, `parseProperty` and `parseRemembers`, since each has its own recovery loop and they have failed independently.

## Why it is worth an issue of its own

[...]

PR #68, opened 2026-09-22
B12: the world root

### LINKED ISSUE #12: B12 · The world root
**Backlog item B12** · Phase 2 — The world model · size M

## What

The world root: one tree, world properties, `pass any (false)` unless written, `visitors are`, `visitors arrive at`, the world composing kinds with `sprout.World` implicit

## Spec

Implements: **The world model; Places** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#3 (B03)

## Blocks

#13 (B13), #14 (B14), #16 (B16)

## Working on this

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B12.

ericeslinger:
Closes #12

The first item of Phase 2. `world printers_shop: victorian.Voice { … }` — the root of the one tree, the only object with no container, the only one that cannot move, and the only one that can be neither spawned nor destroyed.

## Two things about it are load-bearing far beyond this file

Both are here so that nothing downstream has to invent them.

**It refuses to pass.** `pass any (false)` unless it says otherwise is what makes places out of range of one another — so one place hearing another is a deliberate act by the world, not a consequence of sharing a microworld. B15 walks range against that and B32 reads the rules a world writes; `WORLD_PASSES_ANYTHING` is the answer without one.

It is a default of the **language**, not of the host. Every other default in this compiler is a number a host sets; this one is stated in the spec, so it is a constant rather than a limit anybody configures.

**It says what a person is made of.** `visitors are Creature` names the visitor kind — an ordinary kind, so `item.is(sprout.Actor)` stays a nominal test rather than a name the engine knows — and `visitors arrive at composing_room` says where a person begins.

## `sprout.World` is composed whether written or not

```sprout
world printers_shop { … }                    composes sprout.World
world printers_shop: sprout.Voice { … }      composes sprout.World, sprout.Voice
world printers_shop: sprout.World { … }      composes sprout.World — and that is not a collision
```

Writing it explicitly adds nothing rather than colliding. Refusing it would make an implicit rule visible in a way an author cannot act on.

It goes **first** either way, wherever it was written — composition order sequences a composable member's contributions, so leaving a written one where the author put it would make the two spellings mean different things, and "writing it adds nothing" would stop being true the moment `sprout.World` has a member of its own to sequence. That came out of review; see below.

The implicit one is **not** in the declaration's `composes` — that is what was *written*, so a diagnostic about a written kind can point at it. `world.ts` adds the rest.

## What the parser reads, and what it leaves

World members come off a table, the same way `file()` reads declarations, so the message for a word it does not read is built from what reads them and the two cannot drift. Today: properties, `:remembers`, `visitors are`, `visitors arrive at`.

Deliberately **not** here: `contains` and `contains actors` are B13's, which is what makes a place a place; resolving `visitors arrive at` to an object is B14's, since identifier scope does not exist yet; the pass rules themselves are B32's; arrival is B42's. `arriveAt` is kept as the written `Ident`, with its span, for B14.

**One thing cost a round of its own.** Whether a word *is* a member and whether reading it *succeeded* are two questions, and I answered them in one expression:

```ts
: ((readers.get(…) ?? null)?.() ?? undefined);
```

`?? undefined` turns a member that failed into a word nobody knows — so `world w { visitors }` said both *"A world says two things about visitors"* and *"A world is not made of `visitors`."* Split into two, and there is a spec asserting exactly one thing is said.

## Method

**Sixteen deliberate defects** run against the suites, all failing them: `sprout.World` not implicit, or added when already written; a visitor kind or an arrival made optional; either said twice; a duplicate property overwriting the first; a kind composed twice; the world passing everything; a lowercase word read as a kind; `visitors arrive` needing no `at`; a member that failed reported as an unknown word; a broken world swallowing the rest of the file.

**Three of my own mutants were bad and are reported as such** — one a no-op, one with an ambiguous pattern, and one that survived because the code it attacked was *redundant for the inputs the suite had*. That last one turned out to matter: the `recover()` calls are what keep a broken world from being reported twice, so the spec now asserts **exactly one** refusal rather than at least one, and adds the input that exercises it.

## What review changed

Four rounds, eight findings. Every one reproduced against the built parser before being fixed; every fix mutation-tested. Rounds 2 and 3 each found the defect **in the previous round's fix**, and all three blocking findings turned out to be one defect seen from different sides — so the summary is worth as much as the diff.

### Round 1 (`f8da9a5` → `559ffd2`) — 2 blocking, 2 non-blocking

- **A member word that also starts a declaration.** `world w { enum Ward { oak } }` said both *"`w` is never closed"* and *"a world is not made of `enum`"*, orphaning the world's brace and reparsing the enum as a sibling of the world that held it. `worldMembers` now guards at the top of its loop, as `enumDeclaration`'s option loop has since #58.
- **A member whose default value is an unclosed list swallowed the rest of the file.** A read-on-failure guard was not enough: `enum` parses perfectly well *as a list element* — nothing reserves an option's name — so `literal()` took the word and the hunt for a `]` walked on. The guard belongs at the **top** of `listLiteral`'s and `remembers`' loops.
- **The composition list** now names a missing comma instead of letting the brace complain about the wrong problem.
- **`sprout.World` goes first** whether or not it was written.

#66 was strengthened alongside them, because it could see only one of the two. *"Kept, or named in something said"* is satisfied by a misplaced sibling and by a lost name merely being quoted in an unrelated refusal; it now reads **"kept, and nothing appears that was not written"**, with one allowance — a construct that was never closed gives its contents to the file.

### Round 2 (`559ffd2` → `1973486`) — 1 blocking, 1 non-blocking

**Round 1's guard was too blunt and destroyed good input.** It asked only whether a **closer** followed the word. A remembered property is written `name: value`, so the token after a legal entry's name is *always* `:` — so every `:remembers` holding an entry called `enum`, `message` or `world` was read as a declaration starting mid-list and thrown away whole. Inside a world it took the world with it and emitted a stray `An enum needs a name.` naming nothing the author wrote. `[enum silver]` went the same way.

Each declaration word now says what its **own** opening looks like, in `DECLARATION_SHAPES` beside the readers table — `enum Ward {`, `message :stir`, `world name {` / `world name:`, plus the brace alone where a name was forgotten. A word without its opening is just a word. A bare `:` deliberately does **not** count for `world`, though a nameless world is written that way: `[world: 1]` remembers a property called `world`, and when the two readings collide the legal one wins.

`argumentList`'s twin guard was **reverted** — unreachable, and the review verified byte-identical behaviour with and without it.

**#69** is the net that would have caught it, and #66 could not: #66 walks *defective* items and asks what survives beside them, and there was nothing wrong with any of these inputs. #69 is the other half — *nothing well formed is refused at all* — driven off `DECLARATIONS` so `kind` and `object` are covered the day they land. It also replaced three specs that walked `DECLARATIONS` while writing an enum's shape out three times (`${word} Two { a }`, which is not a message) with one `A_DECLARATION` table of real declarations plus a drift check.

### Round 3 (`1973486` → `e1b181b`) — 1 blocking, 1 non-blocking

**Round 2's fix made recovery stop less often, so it ran further and swallowed a second malformed declaration**, telling the author about the first mistake only:

```
enum Ward oak silver
message                      <- never mentioned
enum Glaze { none, shino }
```

The root cause was collapsing **two questions into one predicate**. They have opposite exposures:

- A loop reading elements or members **stops** when the answer is yes, so a false yes throws away what the author wrote — round 2's bug. It wants the question asked **strictly**.
- Recovery is skipping past what it could not read, so a false **no** loses a diagnostic and a false yes costs nothing, because the file's own reader takes the word next and says what is wrong with it — round 3's bug. It wants the question asked **loosely**.

One predicate cannot be both, and each time it moved, one side broke. So there are two now, each carrying the asymmetry that decides it: `atDeclarationStart()` (strict, the shape table, six reading loops) and `atRecoveryStop()` (loose, a faithful restoration of what this file asked before B12, `recover()` and `recoverInBraces()`).

Non-blocking, and it turned into something: no spec reached `world name: Lib` as a boundary, so deleting that clause passed the whole suite. `ALSO_WRITTEN` now carries a word's other spellings and the reading-loop spec walks all of them — and it has to be a *reading-loop* spec, because after the split recovery no longer consults the shape table at all. Splitting then made the forgotten-name clause untested the same way, for the same reason, so that has its own reading-loop spec now.

### Round 4 (`e1b181b` → `a198697`) — no blocking, 1 non-blocking

The split is correctly assigned at all eight call sites, and `atRecoveryStop()` was confirmed identical to the pre-B12 predicate by direct diff against main. But **four of the eight sites had no spec sensitive to which question they asked** — flipping any one of them alone left the whole suite green. Since rounds 2 and 3 were both exactly a call site asking the wrong question and shipping green, that was the gap that mattered.

One input per reading loop, each chosen so the two questions disagree about it — a word this compiler reads, followed by something that is neither that word's own opening nor a closer:

```
enum Ward { oak, message silver }                     -> [oak, message, silver], one "needs a comma"
world w { message foo }                               -> says the word is not a member, before the rest
world w { :x [- message foo] … }                      -> `w` and `Ward` both kept
world outer { :remembers [oops: - message foo] … }    -> `outer` and `Ward` both kept
```

Each pins the full refusal list, not just the surviving declarations, since one of the four loses no content and would otherwise slip. **Flipping each of the eight call sites individually now fails a spec**, where four were silent before.

## One earlier spec had to move

`parse.spec.ts` used `world printers_shop { }` as its example of a word this compiler does *not* read. It reads one now. The example is `kind Vessel { }` (B19's), with a comment saying why it moved — which is the point of building that message from the table rather than writing it out.

## Spec gaps recorded

Three, in the working notes:

- **Whether a world must say what its visitors are and where they arrive.** Every example writes both; the spec never says either is required. Both are refused when missing, on the reading that a world nobody can be made for, or put anywhere in, is not a microworld. Easy to loosen, and the note says what each loosening would need.
- **Whether writing `sprout.World` explicitly is a collision.** It is not, and the reasoning is above.
- **Whether a world's name and its namespace are the same word.** A world's own declarations are unqualified, but nothing states that its *name* is its library name. `resolveWorld` takes the library as a parameter rather than deriving it, so whichever way B21 settles it, this does not have to change.

And two that read like gaps and are not: `pass any (false)` being a language default rather than a host one, and what a world holds being a property like any other — resolved by B06's own resolver, because the world is an object.

## Receipts

Both at `a198697`, the PR head.

```
npm run gate
```
(exit 0 — 1048 passed | 7 skipped across lang/core/stores, 40 passed in cli; 1003 before this PR)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at a198697)
```


ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:839-861` (`worldDeclaration`'s member loop, with `worldMemberReader` at 866-877)

A member word that also spells a top-level declaration (`enum`, `message`, `world`) is reported twice, and the text after it is reparsed by `file()` as an unrelated top-level sibling, orphaning the world's own closing brace.

Verified against the built parser at f8da9a5:
```
world w {
  enum Ward { oak, silver }
}
```
returns `declarations = [enum Ward]` — `world w` never appears at all — with three refusals: "A world is not made of `enum`.", "`w` is never closed.", and `Sprout does not know what to do with "}" here.` The `enum Ward {...}` written inside `w` becomes a sibling top-level declaration; anything else `w` declared goes with it. Same for `message`, and for a nested `world` — including one that is itself well-formed, which survives as a sibling in place of the world that contained it.

Cause: `worldMemberReader()` runs on the current token unconditionally, before anything asks whether that token is about to start a new declaration. `enumDeclaration()`'s own option loop guards against exactly this: it calls `atDeclarationKeyword()` first, before attempting to read an option (parse.ts:394), so an option that happens to spell `enum` or `message` is reported once, as "never closed," and left for `file()` to reparse cleanly. `worldDeclaration()` has no equivalent guard, so it says both.

Not caught by the gate, by the PR's own "does not swallow the declaration written after a broken one" test (its fixtures never use a member word that is itself a declaration keyword), or by the #66 invariant even with `DECLARATION_DEFECTS` extended with a world shape: `nothingVanishes` only checks that a good name is kept somewhere in `declared` or mentioned somewhere in the concatenated diagnostic text, and doesn't notice that what was "kept" is mis-scoped, or that an extra declaration appeared. This will also start firing for `contains`/`kind` and whatever else joins `DECLARATIONS` later.

Fix direction: check `atDeclarationKeyword()` before calling `worldMemberReader()`, the way `enumDeclaration()`'s loop does.

---

ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:817-863`, via `property()`/`remembers()` (953-1025) into `listLiteral()` (634-699, unchanged by this PR)

A world with an unclosed list-valued property silently swallows every declaration written after it, to the end of the file.

Verified against the built parser at f8da9a5:
```
world w { :x [
}
enum Ward { oak }
```
returns `declarations = []` — `enum Ward` is gone, not merely misplaced — with six refusals, four of which blame `}`, `Ward`, `{` and `}` in turn as bad list elements (`` `Ward`, which starts with a capital is not a value. ``), then "This list is never closed." and "`w` is never closed."

Cause: `listLiteral()`'s recovery reads one token at a time hunting for `]`, with no `atDeclarationKeyword()` exit, because before this PR it was reachable only from the standalone `parseProperty`/`parseRemembers` entry points, where running to end-of-file is harmless — nothing else was waiting in `file()`'s declaration list behind it. `worldDeclaration()` is the first construct to put a property inside a brace body that `file()` is still reading, and `listLiteral()`'s recovery was never given the boundary that now needs.

This is the exact category the PR body names as one of its sixteen tested mutants ("a broken world swallowing the rest of the file") and that #66 is meant to stand behind, and it still reproduces because neither exercises an unclosed list. It would also survive `DECLARATION_DEFECTS` being extended with this shape, for the same reason as the previous comment: `Ward` shows up in the diagnostic text (as the thing being called an invalid value), so `nothingVanishes` reads it as present even though the declaration is gone.

Fix direction: give `listLiteral()` (and the other read-on-failure loops) an exit at `atDeclarationKeyword()`, or have `worldDeclaration()` bound how far a member's own recovery may run past the tokens it owns.

---

ericeslinger:
**Non-blocking** — `sprout/lang/src/parse.ts:794-805` (the composition list in `worldDeclaration`)

The composition list has no "missing comma" detection, unlike every other comma-separated list in this file (`remembers`, list literals, argument lists all defer a "needs a comma" diagnostic via `separator()`). It just stops reading kinds the moment a comma isn't there.

`world w: victorian.Voice OtherLib.Kind { visitors are P\n visitors arrive at y }` refuses only with "`w` has nothing in it." / "A world is written `world <name> { … }`, holding what it is made of." — the wrong problem; the real one is the missing comma between the two composed kinds. Recovery is otherwise clean (nothing after is swallowed), so this is a diagnostic-quality gap, not a correctness one.

---

ericeslinger:
**Non-blocking** — `sprout/lang/src/world.ts:81-123`

Composing `sprout.World` implicitly always puts it first in `composes`; writing it explicitly leaves it wherever it was written. Verified: `world w: victorian.Voice { … }` resolves to `[sprout.World, victorian.Voice]`, and `world w: victorian.Voice, sprout.World { … }` resolves to `[victorian.Voice, sprout.World]`. The spec says composition order "sequences effects" for composable members (handlers, default passages, …), so the two spellings aren't quite interchangeable once `sprout.World` has a composable member of its own — which sits a little awkwardly next to the working notes' "[writing it explicitly] adds nothing." Not observable yet (nothing reads composition order until B19/the standard library exist), but worth deciding before B19 builds on this shape: force `sprout.World` to the front regardless of where it's written, or say plainly its position is significant.

---

ericeslinger:
**Review**: 2 blocking, 2 non-blocking. Gate at f8da9a5: green, reproduced independently (sprout 1032 passed | 7 skipped, cli 40 passed, exit 0 — matches the PR's receipt exactly). e2e not re-run: the diff is confined to `sprout/lang/src` (parser, AST, binder), no `cli/`, `sprout/core`, packaging or example-world changes, so the author's e2e receipt stands. `world.ts`'s reading of the spec (implicit `sprout.World`, the `visitors are`/`arrive at` requirement, `pass any (false)`, deferred `contains`) checks out, and its open questions are properly recorded as holes in the working notes. Both blocking findings are recovery gaps in the new `worldDeclaration`/`worldMemberReader` machinery — a member word that doubles as a top-level keyword, and a member whose default value is an unclosed list — that let a broken world either duplicate its own diagnostic and misplace a nested declaration, or silently swallow a well-formed one after it. Both reproduce against the built parser and both slip past the PR's own targeted tests and the #66 invariant.

---

ericeslinger:
All four fixed at `559ffd2`. I reproduced each one against the built parser before changing anything, and each fix is mutation-tested: reverting any one of the five changes below fails a spec.

**Blocking 1 — a member word that also starts a declaration.** Right, and right about the cause. `worldMembers` now guards at the top of its loop, the way `enumDeclaration`'s option loop has since #58:

```ts
if (this.atDeclarationKeyword()) {
  this.diagnostics.refuse(
    this.peek().at,
    `\`${name.text}\` is never closed.`,
    'Add a } after what the world is made of.',
  );
  return null;
}
```

`world w { enum Ward { oak } }` now says `` `w` is never closed. `` once, and `Ward` is the file's enum. Also covered for `message` and for a nested `world`.

**Blocking 2 — an unclosed list swallowing the file.** Fixed, but your fix direction was half of it, and the missing half is the interesting part. I put the exit on the read-on branch first, and it did nothing: `enum` **parses perfectly well as a list element**. Nothing reserves an option's name, so `literal()` takes the word, `listLiteral` never reaches its failure branch, and the hunt for a `]` walks on through the file exactly as before. The guard has to be at the **top** of the loop, before the element is read — which `remembers` needed too, for the same reason.

Which makes the other half of your sentence load-bearing: the word is only a boundary because of what *follows* it. `atDeclarationKeyword()` asks that question, so `[oak, enum]` and `[oak, message]` are still lists of two options. There is a spec for that, because it is the thing a too-eager fix here breaks.

**Non-blocking 3 — the missing comma.** Added, via an `atKindName()` that distinguishes "another kind was written" from "the list ended". `world w: victorian.Voice other.Kind { … }` now says `A world needs a comma between the kinds it composes.` and reads both kinds.

**Non-blocking 4 — where a written `sprout.World` goes.** Taking the first of your two options: it is forced to the front regardless of where it is written. Your argument is the right one — if position is significant then "writing it adds nothing" is already false, and it becomes properly false the moment `sprout.World` has a composable member to sequence. Better to settle it now than to have B19 discover it. The working notes' hole is unchanged in substance; the code comment now says why the order is not the author's to choose.

**And the invariant itself.** You were right that `nothingVanishes` could not see either finding, and right about why: "kept **or** named in something said" is satisfied by a misplaced sibling and by a lost name merely being quoted in an unrelated "is not a value" refusal. It now reads

> kept — **and** nothing appears that was not written

with one deliberate allowance: a construct that was never closed gives its contents to the file, so a name written *inside* the defect may legitimately surface at the top level. That is the rule enums have followed since #58 and it is what keeps the author's work instead of skipping to a brace and losing it. A name that was never written anywhere still may not appear. Seven world shapes join `DECLARATION_DEFECTS`; `world w { }` is deliberately not one of them, with a comment saying why — it parses, and what is wrong with it is `resolveWorld`'s to say.

Strengthened, the invariant catches blocking 2 on its own. It does not catch blocking 1 (a duplicated diagnostic loses nothing), nor 3 or 4, so all four have direct specs as well — verified by reverting each fix in turn.

Receipts in the body updated to `559ffd2`. Gate exit 0, e2e green.

---

ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:1043` (`remembers()`'s loop) and `parse.ts:650` (`listLiteral()`'s loop), via `atDeclarationKeyword()` (`parse.ts:213-219`)

The guard added to close the previous round's blocking-2 is too blunt: `atDeclarationKeyword()` treats a word as starting a declaration whenever it is *not* immediately followed by `,`/`}`/`]`/end, which is the wrong test for `remembers` — a remembered property's grammar is `name: value`, so the token after a legal entry's name is always `:`, never a closer. Any `:remembers` entry literally named `enum`, `message` or `world` is now unparseable, unconditionally, with no typo required.

Verified against the built parser. At `559ffd2`:
```
:remembers [enum: 1]
```
→ `parseRemembers` returns `null`, one refusal: "This `:remembers` is never closed." — it closed fine; the property is simply named `enum`. At `f8da9a5` (one commit earlier) this parsed cleanly, zero refusals, `{name: 'enum', default: 1}`. Same loss for `message:`/`world:`, in any position — `:remembers [foo: 1, enum: 2]` loses the whole block, not just the tail.

The list-literal side needs one more ingredient — a missing comma right after the word — but is the same defect: `:x [enum silver]` (two option names, comma dropped) now says "This list is never closed." and returns `null`; at `f8da9a5` it read as two elements plus "A list needs a comma between its elements." `:x [enum 5]` and `:x [oak, enum silver]` fail the same new way. (`[oak, enum]`, `[enum, oak]`, `[enum]` are all still fine — the reply's own reasoning holds whenever the word actually is followed by a closer.)

Inside a world this reproduces the exact class of bug this round's PR claims closed, just via a different door:
```
world w {
  :remembers [enum: 1]
  visitors are Creature
  visitors arrive at start
}
enum Ward { oak, silver }
```
→ `world w` is entirely absent from `declarations` — not misplaced, gone — with refusals "This `:remembers` is never closed.", "`w` is never closed.", and a bogus "An enum needs a name." that names nothing the author wrote wrong: the failed guards never consume the `enum` token, `worldMembers`' own new guard (line 867) fires on the same token next, and `file()` then recovers onto that leftover `enum` and tries to read it as a real declaration before recovering a second time onto the true `Ward`. `:x [enum silver]` in a world body loses the world the same way. Neither shape is in `ENTRY_DEFECTS`, `ELEMENT_DEFECTS` or `DECLARATION_DEFECTS`, so the strengthened #66 invariant doesn't see it, and this round's two new specs (`':x ['`, `':remembers [a: 0'`) test the *unclosed* case, not the *closed-but-keyword-named* one.

Fix direction: `atDeclarationKeyword()` needs to check what a declaration's name actually looks like (a capitalized `kind` token) rather than inferring "not a declaration" only from the absence of a closer — or `remembers`' loop needs a narrower check than the one written for bare list elements, since an entry's grammar has a mandatory `:` that an element does not.

---

ericeslinger:
**Non-blocking** — `sprout/lang/src/parse.ts:1371` (`argumentList`'s read-on branch)

Appears to be a no-op given the current grammar, and it isn't tested. `argumentList` only got the guard on the "argument failed to read" branch (not at the top of the loop, unlike `listLiteral`/`remembers`). `primary()` (parse.ts:1245) and `postfix()`'s member-name read both accept *any* bare `name` token unconditionally, including `enum`/`message`/`world` — so `expression()` never returns `null` while positioned on a token where `atDeclarationKeyword()` could be true. Verified across 28 constructed inputs (`self.set(enum)`, `self.set(enum 1)`, `self.set(enum.message(1, world))`, `self.set(-enum)`, `self.set(:wear.enum)`, and more): diagnostics and results are byte-identical between `f8da9a5` and `559ffd2` for every one. No new spec exercises this branch either. Not urgent since it changes nothing today, but it leans on the same heuristic just shown to be unsound above — worth a test if there's a real path to `argument === null` landing on a reader-keyword token, or dropping the change until there is one.

---

ericeslinger:
**Review (round 2, `f8da9a5..559ffd2`)**: 1 blocking, 1 non-blocking. Gate at 559ffd2: green (1037 passed | 7 skipped sprout, 40 passed cli, exit 0 — matches the receipt). The two fixes from round 1 are correct for the shapes that were tested, but the shared `atDeclarationKeyword()` guard they both lean on is too blunt: it now silently discards any `:remembers` entry named `enum`/`message`/`world` outright, and a list missing a comma right after such a word, reproducing "a world vanishes" — the exact failure class this round closed — through a different door. `argumentList`'s twin guard, `worldMembers`' own guard, the strengthened #66 invariant, and `world.ts`'s composition ordering all check out on direct testing.

The other three priorities, checked with nothing to report:

- **`worldMembers`' own guard (parse.ts:867).** In isolation it's fine: it only inspects the member's *leading* token, and property/`:remembers` symbols and the bare word `visitors` never match `atDeclarationKeyword()` (symbols aren't `name`-kind; `visitors` isn't a reader). The bad interaction is downstream, folded into the Blocking finding above: when `remembers()`/`listLiteral()` fail without consuming the offending token, this guard re-fires on that same token and compounds the loss. A world whose last member is well-formed parses fine (checked directly, zero refusals).
- **The widened #66 `written` allowance, `defect.match(/[A-Z][A-Za-z_]*/g)`.** Not too wide for what's in the table today: it only admits capitalized words that are lexically present in the literal `defect` string being tested, and the parser cannot manufacture a name from text that isn't there, so it can't mask a fabricated declaration for the existing entries. It simply doesn't cover the shapes in the Blocking finding, because those shapes (`:remembers [enum: 1]`, `:x [enum silver]`) aren't in `ENTRY_DEFECTS`/`ELEMENT_DEFECTS`/`DECLARATION_DEFECTS` — which is exactly why the gate stayed green.
- **`world.ts`'s `ordered`.** Checked directly against the built resolver, not by hand: `sprout.World, sprout.World` → refused once ("composes `sprout.World` twice"), `composes` ends up `['sprout.World']` (no duplicate). `sprout.World, victorian.Voice, sprout.World` → same refusal, `composes` = `['sprout.World', 'victorian.Voice']` (Voice kept). `victorian.Voice, victorian.Voice` → refused for Voice, `composes` = `['sprout.World', 'victorian.Voice']` (World still implicit-first, nothing lost). No way found to make a world compose `sprout.World` twice or lose a kind.

All findings above were reproduced against the built parser (vitest, real imports, not hand-traced), both at `559ffd2` and, for contrast, at `f8da9a5`.

---

ericeslinger:
Both fixed at `1973486`. Reproduced first, exactly as reported:

```
:remembers [enum: 1]  -> null,  ["This `:remembers` is never closed."]
:x [enum silver]      -> null,  ["This list is never closed."]
world w { :remembers [enum: 1] … }
                      -> ['Ward'],  [… "`w` is never closed.", "An enum needs a name."]
```

This is the eleventh regression of this shape in `parse.ts`, and unlike most of the ten it is not one I inherited — I wrote it, in a fix for the last round's finding, and shipped it behind a green gate. Worth saying plainly.

**The guard.** Your diagnosis is exact and so is the fix direction. Each declaration word now says what its *own* opening looks like, in a `DECLARATION_SHAPES` table beside the readers table:

```
enum Ward {          message :stir          world name {   |   world name:
```

A word without its opening is just a word. Two tokens is as far as it looks, deliberately — a guard that asks for more starts refusing declarations an author really did write.

Two judgement calls came out of it, both now written down at the point they are made:

- **The brace alone counts as an opening.** Requiring `kind` + `{` for `enum` broke `enum { a }\nenum { b }\nenum { c }` down to one refusal instead of three — an author who forgot the name still started a declaration. So `enum {` and `world {` are boundaries.
- **A bare `:` does not count, for `world`.** A nameless world *is* written `world: victorian.Voice { … }`, so admitting it is tempting and symmetric. But `[world: 1]` remembers a property called `world`, and that is input an author meant where a world with no name is input they did not. When the two readings collide, the legal one wins.

**`argumentList`.** You're right that it's unreachable and untested, so I've reverted it rather than adding a spec for a branch nothing can enter. Unreachable code with no spec is how the other ten of these hid; if a path to `argument === null` on a reader word ever appears, it arrives with the spec that found it.

**The net.** #66 genuinely could not see this, and for the reason you gave: it walks *defective* items and asks what survives beside them, and there was nothing wrong with any of these inputs. So there is a companion invariant, #69 — *a declaration's own word is an ordinary name wherever a name may stand*, i.e. nothing well formed is refused at all. It is driven off `DECLARATIONS`, so `kind` and `object` are covered the day they land rather than the day somebody remembers:

- as the name of a remembered property,
- as an option in a list, in any position,
- with a dropped comma after it — which is *a missing comma*, not the end of the list,
- and inside a world, which is where losing it cost the whole declaration.

Running it against the exact code that shipped in `559ffd2` fails three of its four cases.

That also turned up three specs that walked `DECLARATIONS` while writing an enum's shape out three times — `${word} Two { a }`, which is not a message. They now share one `A_DECLARATION` table of real declarations, with a drift check on it.

**Mutants: six, all caught.** The blunt guard as shipped; `enum` losing its forgotten-name case; `world` admitting a bare `:`; `message` admitting any name; and round 1's two fixes re-reverted, to be sure tightening the guard had not quietly reopened them.

Receipts at `1973486`: gate exit 0 (1041 passed | 7 skipped sprout, 40 cli), e2e green. Body updated.

---

ericeslinger:
**Blocking** — `sprout/lang/src/parse.ts:333` (`recover()`), also via `sprout/lang/src/parse.ts:363` (`recoverInBraces()`)

Recovery now silently drops the diagnostic for a second, nearby malformed declaration attempt, because `message`'s shape (`parse.ts:100`) has no fallback at all and `world`'s shape (`parse.ts:108-112`) deliberately excludes a bare `:` — and `recover()`/`recoverInBraces()` treat anything their shape doesn't recognize as ordinary filler, with nothing else left to report it.

Input:
```
enum Ward oak silver
message
enum Glaze { none, shino }
```
At this PR's head (1973486), `parseDeclarations` reports only `"The options of \`Ward\` go in braces."` — the bare `message` on line 2 (itself a mistake — a forgotten name) is never mentioned anywhere. `recover()`, hunting from `Ward`'s failure for the next declaration, checks `message`'s shape (needs a symbol token right after; `enum` doesn't qualify), decides it isn't an opening, and just steps over the whole word as filler, landing on `enum Glaze`. At `559ffd2` *and* at main (`a34a6d8`) — where `message` has always been a declaration keyword — the identical input produces a second refusal, `"A message needs a name."` I ran all three commits to confirm; only `1973486` drops it.

Same mechanism through `recoverInBraces()`, for `world`'s excluded bare `:`: `enum Ward { oak,\n42\nworld: v.X {\n  contains\n}\nenum Glaze { none, shino }` gives only the two `Ward` refusals at `1973486` — the malformed `world: v.X { contains }`, braces and all, is absorbed whole into `Ward`'s "never closed" recovery (its own `{`/`}` net out the brace-depth count) with no diagnostic of its own. `559ffd2` additionally reports `"A world needs a name."` for it.

This regresses exactly what both earlier rounds of this review were about, and this file's own header comment ("an author owed three problems is owed all three"). A fix that keeps the round-2 correction (word-as-ordinary-name) but adds a secondary stop when the word is clearly not an ordinary-name use either — not immediately followed by `:`, `,`, `}`, `]`, or end — would close this without reopening round 2's bug.

---

ericeslinger:
**Non-blocking** — `sprout/lang/src/parse.ts:111` (`world`'s `DECLARATION_SHAPES` entry) / `sprout/lang/src/parse.spec.ts`

No spec exercises `world name: Lib {` (the composing form) as a recovery boundary — every test that loops `DECLARATIONS` (the "stops recovery…" spec at `parse.spec.ts:787`, and all four `#69` cases) uses `A_DECLARATION.world`, which is brace-only (`world two { visitors are P … }`). I deleted `|| punct(after, ':')` from world's shape and reran: `parse.spec.ts` still passes 169/169, and the full `sprout` suite still passes 1041/1041. The shipped code is correct — I confirmed with the real implementation that recovery does stop at `world foo: Lib {` mid-file — so nothing ships broken, but the commit message's "six mutants, all caught" claim doesn't cover this one; a seventh (dropping just the colon clause) is silent. A recovery-boundary case using the colon form would close the gap.

---

ericeslinger:
**Review (round 3, `559ffd2..1973486`)**: 1 blocking, 1 non-blocking. Gate at 1973486: green (exit 0; sprout 1041 passed | 7 skipped, cli 40 passed — matches the PR's receipt). `DECLARATION_SHAPES` correctly fixes round 2's false positives (word-as-ordinary-name now works), but `message`'s no-fallback shape and `world`'s bare-`:` exclusion leave `recover()`/`recoverInBraces()` silently swallowing a *second*, nearby, genuinely malformed declaration instead of reporting it — a narrower instance of the same bug class both earlier rounds caught.

Checked and found clean, no finding filed:

- `atDeclarationKeyword()`'s narrower shapes never produce a false negative on *well-formed* enum/message/world syntax — `enumDeclaration()`/`messageDeclaration()`/`worldDeclaration()`'s own success conditions are exactly mirrored, so no legitimately-written declaration is lost, only the diagnostic for a malformed one (above).
- `argumentList`'s reverted guard: confirmed dead by reading `primary()` — a bare reader-keyword name is always consumed successfully as a binding/kind-expr/free-call before `expression()` can return null on it, so the guard was unreachable; the revert is safe.
- The `A_DECLARATION`/`sampleOf()` refactor of the three pre-existing specs: not weakened. The old `${word} Two { a }` template was actually *wrong* for `message` under the new precise shapes (`message Two { a }` was never a real message opening); the new per-word real samples assert the same boundary behaviour more accurately. `sampleOf()`'s drift check is real (`expect(...).toContain(word)` throws if a table falls behind `DECLARATIONS`).
- Bare `{` misfiring on legal non-declaration uses of the word: not currently reachable — `{` is used nowhere else in this parser's grammar yet (only enum bodies and world bodies), verified by grepping every `punct(..., '{')`/`take('punct','{')` site.
- `skipBracketed()` does not call `atDeclarationKeyword()` in either commit and is unaffected by this diff (its callers are all nesting-cap overflow / bracket-mismatch paths, unrelated to declaration recovery).

All parser behaviour above was checked by running `parseDeclarations` at each of `1973486`, `559ffd2` and `a34a6d8` in its own worktree, not hand-traced.

---

ericeslinger:
Both fixed at `e1b181b`. Reproduced first, exactly as reported — only `"The options of \`Ward\` go in braces."`, with the bare `message` never mentioned.

**The root cause is one I should have seen when I wrote the tightening**, and your finding names it precisely: I had collapsed two questions into one predicate, and they have *opposite* exposures.

- A loop reading elements or members **stops** when the answer is yes, so a false yes throws away what the author wrote. That is round 2's bug. It wants the question asked **strictly**.
- Recovery is skipping past what it could not read, so a false **no** loses a diagnostic, and a false yes costs almost nothing — the file's own reader takes the word next and says what is wrong with it. It wants the question asked **loosely**.

One predicate cannot be both, and every time I moved it, one side broke. So there are now two, each carrying the asymmetry that decides it:

- `atDeclarationStart()` — strict, `DECLARATION_SHAPES`, for the six reading loops.
- `atRecoveryStop()` — loose, exactly what this file asked before B12, for `recover()` and `recoverInBraces()`.

That is your suggested fix, arrived at from the other end: rather than adding a secondary stop to the strict question, the loose question goes back where it was always right. `CLOSERS` comes back with it.

**Non-blocking, and it turned into something.** You were right that no spec reached `world name: Lib` — `ALSO_WRITTEN` now carries a word's other spellings and the reading-loop spec walks all of them. But note where that spec had to go: a *recovery* test could not have caught it after the split, because recovery no longer consults the shape table at all. Your mutant was silent for one reason at `1973486` and would have been silent for a different one at `e1b181b`.

Splitting the predicate then made the forgotten-name clause (`enum {`, `world {`) untested in the same way — the spec that covered it, `enum { a }\nenum { b }\nenum { c }`, goes through recovery. I only found that because the mutant flipped from caught to surviving between two runs an hour apart. It has a reading-loop spec of its own now.

**Eight mutants, all caught**: each predicate given the other's question; each clause of the shape table deleted in turn (`enum`'s bare brace, `message`'s symbol, `world`'s colon); and round 1's two fixes re-reverted.

Receipts at `e1b181b`: gate exit 0 (1044 passed | 7 skipped sprout, 40 cli), e2e green. Body updated.

Three rounds, and each one found a real defect in the previous round's fix — all three in the same few lines of recovery. I don't think a fourth round on this diff is the right use of it: what the three rounds actually converged on is that the two questions were conflated, which is now fixed at the root rather than patched at a call site, and the eight mutants pin both halves. Happy to be told otherwise.

---

ericeslinger:
**Non-blocking** — `sprout/lang/src/parse.ts:473` (`enumDeclaration`'s option loop), `:763` and `:1139` (`listLiteral`'s and `remembers`' post-failure branches), `:946` (`worldMembers`' loop guard)

Four of the eight call sites this round assigns a predicate to have no spec sensitive to which predicate they get. Flipping any *one* of these four, alone, from `atDeclarationStart()` (strict, as shipped) to `atRecoveryStop()` (loose — the wrong choice for a reading loop, by the PR's own asymmetry argument) leaves all 172 tests in `parse.spec.ts` passing.

Verified against the built parser (one-line mutation, full suite re-run, restored before the next):
- Flip only line 473: `enum Ward { oak, message silver }` goes from `Ward = [oak, message, silver]` + one "needs a comma" (current, correct) to `Ward = [oak]` — `silver` vanishes from the output entirely, quoted in no diagnostic, and `message` is retried and fails as a stray top-level declaration. Suite: 172/172 still pass.
- Flip only line 763 or only line 1139: `world outer {\n  :remembers [oops: - message foo]\n  visitors are Creature\n  visitors arrive at start\n}\nenum Ward { oak }\n` goes from `outer` kept whole with three precise diagnostics (current) to `outer` lost entirely ("`outer` is never closed.") — the exact shape of bug rounds 1 and 3 fixed. (`:x [- message foo]` in place of `:remembers` reproduces the same loss via 763.) Suite: 172/172 still pass.
- Flip only line 946: milder — suppresses "A world is not made of `message`." without losing content. Suite: 172/172 still pass.

By contrast, flipping line 366 or 394 (`recover`/`recoverInBraces`) or 729 or 1122 (the other two reading-loop sites) each fails 1–2 tests — the suite does catch a wrong assignment there, mostly via the `#69` describe block.

The shipped choice is right at all four sites — I could not construct an input where it produces worse output than the alternative, only ones where the alternative is worse. So this is a coverage gap, not a live bug. But given this file's history (eleven-plus regressions in exactly this code) and this round's own "eight mutants, all caught" framing, a future change re-flipping one of these four would ship behind a green gate the same way round 2's and round 3's bugs did. Worth a spec driven off the eight call sites (or the strict/loose split itself) rather than off `DECLARATIONS` alone.

---

ericeslinger:
**Review (round 4, `1973486..e1b181b`)**: no blocking findings, 1 non-blocking. Gate at `e1b181b`: green (exit 0; sprout 1044 passed | 7 skipped, cli 40 passed — reproduced twice independently, matches the PR's receipt exactly). The split (`atDeclarationStart` strict for the six reading loops, `atRecoveryStop` loose for `recover()`/`recoverInBraces()`) is correctly assigned at all eight call sites — verified by running the built parser and mutating each site individually, not by hand-tracing — and `atRecoveryStop()` is a faithful restoration of pre-B12 behavior. One non-blocking finding: four of those eight sites (473, 763, 946, 1139) have no spec that would catch a wrong assignment there, even though the shipped assignment is in fact correct at each.

Checked and found clean, no finding filed:

- **`atRecoveryStop()` = pre-B12 `atDeclarationKeyword()`, exactly**: confirmed identical function bodies against main (`a34a6d8`) by direct diff, and confirmed behaviourally — the only observable difference against main on any input is that `world` is now a registered reader (this PR's own point), not a change in the predicate's logic.
- **Loose recovery over-firing / `recoverInBraces` at wrong depth**: the classic case (`enum Ward { 4 { message :x } }`, an unmatched nested brace) is already covered by an existing spec ("does not promote what is nested inside an unmatched brace to the top of the file") and is unaffected by this round's diff — the depth-guard that prevents promotion (`depth === 1 && …`) wasn't touched. Constructed several brace-depth and dash-then-keyword cases; found no new duplicated/misplaced/nonsensical diagnostic and no new promotion.
- **Compounding between the split questions**: tested several "two malformed declarations near each other" combinations, including ones that only diverge because of the split. No case produced two diagnostics about one word, a word consumed twice, or neither — the handoff from a reading loop's strict "no" into `recover()`'s loose "yes" lands cleanly every time I tried.
- **The new specs assert what they claim**: `ALSO_WRITTEN`/`spellingsOf()` catches deleting `world`'s `:` clause; the new forgotten-name reading-loop spec catches deleting `enum`'s bare-`{` clause.
- Round 1's, round 2's and round 3's specific bugs (nested `enum`/`message`/`world` orphaning the world's brace; the blunt guard eating `:remembers [enum: 1]`/`[enum silver]`; `message`/`world`-with-forgotten-name going silent in recovery) all remain fixed — re-ran the exact inputs from each round's finding against `e1b181b` directly.

Built a throwaway worktree holding `parse.ts` from `a34a6d8`, `1973486`, `559ffd2` and `e1b181b` side by side and ran dozens of constructed inputs through all four via real `parseDeclarations()` calls — never hand-traced, after an early hand-trace of mine turned out to need correcting once I actually ran it.

---

ericeslinger:
Closed at `a198697`, and this one is worth taking seriously even though nothing was broken.

You're right that "eight mutants, all caught" was a weaker claim than it sounded. My eight attacked the *shape table* and the two predicates wholesale; none of them asked the question you asked, which is whether each individual call site is *held* to the question it was given. Four were not. Given that rounds 2 and 3 were both exactly a call site asking the wrong question and shipping green, that gap is the one that matters most here.

Your four discriminating inputs are now specs, verbatim in substance:

```
enum Ward { oak, message silver }      -> [oak, message, silver], one "needs a comma"
world w { message foo }                -> says the word is not a member, before the rest
world w { :x [- message foo] … }       -> `w` and `Ward` both kept
world outer { :remembers [oops: - message foo] … }   -> `outer` and `Ward` both kept
```

Each pins the full refusal list rather than just the surviving declarations, since the milder one (946) loses no content and would otherwise slip.

Then I ran your experiment rather than assuming: flip each of the eight call sites individually, one line, full suite. **All eight now fail a spec**, where four were silent before — 366 and 394 the other way round, to the strict question.

I'm calling the PR ready here rather than going to a fifth round. Four rounds found seven findings, and the three blocking ones were all the same defect seen from different sides — one predicate serving two callers with opposite exposures. That is fixed at the root now, and this round's finding was about holding it there, which is exactly what was missing. If you'd rather it kept going, say so and I'll keep going.

Receipts at `a198697`: gate exit 0 (1048 passed | 7 skipped sprout, 40 cli), e2e green.

---

Issue #69, opened 2026-09-22
Invariant: a declaration's own word is an ordinary name wherever a name may stand


ericeslinger:
Companion to #66. That one says *a defect in one item never loses a well-formed neighbour*; this one says the other half — **nothing well formed is refused at all**.

## Why it needed saying

*Reserved names* reserves nothing for a property's name or an option's, so `enum`, `message` and `world` are ordinary words almost everywhere:

```sprout
:remembers [enum: 1, keep: 2]
:x [Ward] default [oak, enum, silver]
```

Both are legal. Recovery, though, has to tell a declaration from a word that merely spells one, and the first guard written for that (in the review of #68) asked only whether a **closer** followed the word. A remembered property is written `name: value`, so the token after a legal entry's name is *always* `:` and never a closer — which meant every `:remembers` holding an entry called `enum` was read as a declaration starting mid-list and thrown away whole. Inside a world it took the world with it, and emitted a stray `An enum needs a name.` naming nothing the author wrote. `[enum silver]` — two options, one missing comma — went the same way.

#66 could not see any of it: #66 walks *defective* items and asks what survives beside them, and there was nothing wrong with these.

## The rule

For every word in `DECLARATIONS`, that word parses cleanly as

- the name of a remembered property,
- an option in a list, in any position,
- an option followed by a dropped comma — which is *a missing comma*, not the end of the list,

and a world holding any of the above survives, with the declaration after it.

Driven off `DECLARATIONS` rather than written out, so `kind` and `object` are covered the day they are added rather than the day somebody remembers.

## What it protects

`DECLARATION_SHAPES` in `parse.ts` — each declaration word says what its own opening looks like (`enum Ward {`, `message :stir`, `world name {` / `world name:`), plus the brace alone where a name was forgotten. A word without its own opening is just a word. `world` deliberately does **not** admit a bare `:`, because `[world: 1]` is input an author meant where a nameless `world: Voice { … }` is input they did not.

Two later rounds of #68's review turned that into a second, related rule, which lives beside this one in `parse.spec.ts`: **recovery and reading ask the same word different questions**, and must be allowed to be wrong in opposite directions. A loop reading what the author wrote stops on a yes, so a false yes loses their work (this issue). Recovery is already skipping, so a false *no* loses their diagnostic. Hence `atDeclarationStart` (strict, this table) and `atRecoveryStop` (loose), with a spec per call site holding each to its own question.

## State

[...]

PR #70, opened 2026-09-22
B13: containment is a declaration

### LINKED ISSUE #13: B13 · `contains` and `contains actors` as declared capabilities
**Backlog item B13** · Phase 2 — The world model · size S

## What

`contains` and `contains actors` as declared capabilities; a place is whatever declares the latter; exits only on places

## Spec

Implements: **The world model › Places; Containment is a declaration** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#12 (B12)

## Blocks

#15 (B15), #17 (B17)

## Working on this

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B13.

ericeslinger:
Closes #13

`contains` is the primitive and `contains actors` is the capability beside it, and declaring the second is the whole of what makes a place a place. There are no rooms — a wardrobe that declares it can be entered, and everything that follows from being somewhere follows from that one line.

## Why these are declarations and not guards

Neither is a kind the engine knows by name, which is what makes a library's container and the standard library's equally real. And neither can be a guard, because the distinction is **structural**: the engine has to know whether a thing holds others in order to build the tree at all, and no guard can answer that. The spec's *Portability is a guard, and the default belongs in a library kind* makes the same cut from the other side — `contains` is structure, portability is policy, and policy is what guards are for.

A world is the only body that can hold one today, so that is where they are read.

## `KindRef` finally means something

`KindRef.contains` has carried a comment since B08 saying *"what declares it is B13's; typing only asks"*. It now says what it means — **may hold others**, whichever line said so — and has `containsActors` beside it, which is what B15 walks range against and what B17 finds an actor's place by, rather than either of them deciding what a place is. B19 is what will build one from a kind as written; until then a world is the only source.

## Three judgement calls, all recorded

Each is a line in the working notes' *Holes in the spec*.

**`contains actors` implies holding.** The spec calls it "a declared capability beside `contains`" — which reads as two separate lines an author may write — and never says whether the second gives the first. The standard library's own `kind Place` declares only `contains actors`, and `composing_room` holds a bench, a shelf and a kettle. So it implies it. The other reading would make `kind Place` as the spec writes it a place that cannot hold a bench, and would make every place in every world write two lines.

**Writing either twice says nothing.** *How members combine* calls `contains` and `contains actors` idempotent, which settles what happens when **composition** brings two of them together and says nothing about one body writing the same line twice. Nothing is said about it here: a line the table calls idempotent does not become a mistake for being written by one author rather than two — where a property declared twice in one body *is* refused, because the second one would have to win. A redundant one may deserve a warning, and the list of warnings is B50's.

**A world may declare `contains actors`, and so be a place itself.** The world "holds what belongs to no single place", which reads as a thing that is not one — but nothing forbids the line, and *Places inside places* is happy for a place to sit inside another. Accepted; refusing it would be the larger change and would need a sentence in the spec to point at.

### And one that reads like a gap and is not

**A world is not refused for holding nothing.** Every world in the spec writes `contains`, so one without it looks like a world that forgot. It is not: `sprout.World` declares `contains` itself and every world composes it, so the capability arrives through composition, which is B19's to merge. `ResolvedWorld.contains` is therefore what *this declaration wrote*, not what the world ends up with.

## Deliberately not here

**Exits only on places** — the third clause of the issue. Exits are B28's, which is where they arrive at all; this is what B28 will ask. `visitors arrive at` resolving to an object is B14's, which now has `containsActors` to ask whether what it resolved to is a place.

## One thing the parser had to get right

A word after `contains` that is not `actors` is **left where the author wrote it** rather than swallowed. `contains actor`, singular, goes back to the member table, which names `actor` — the part they got wrong — instead of pointing the refusal at `contains`, the part they got right. There is a spec for it.

## Method

**Eight deliberate defects**, all failing the suites: `contains actors` no longer implying holding; it read as a plain `contains`; a second `contains` overwriting what the first said; a plain `contains` forgotten where `contains actors` is not written; a world that writes nothing holding everything; the parser dropping the `actors`; the parser swallowing whatever word follows; the span covering only the keyword.

**One of them found a real problem in my first cut.** I wrote `contains: holds || holdsActors` in the resolver, and the mutant that deleted the `|| holdsActors` survived — because `holds` was already set by *any* `contains` member, so the implication was not expressed anywhere a change could break it. Each flag is now one line the author actually wrote (`wroteContains`, `wroteContainsActors`), with the implication between them in a single place, and the mutant fails.

A suite also caught a mistake in my own spec: I asserted `contains` twice made a place. It does not.

## Review

One round, no blocking findings, one non-blocking: adding `contains()` anchored on `private visitors()` and pushed that method's one-line doc comment above the new one, leaving `visitors()` undocumented. Fixed at `3d2aa5d`.

Worth recording what the review did beyond that, since both are the kind of check that would have saved #68 three rounds. It confirmed the new member word cannot disturb B12's `atDeclarationStart`/`atRecoveryStop` split not by arguing from the structure but by finding that an existing `#69` spec already carries a literal `contains` token through a nameless-world recovery block, and checking it still passes. And it checked this PR body's claim that `sprout.World` declares `contains` against the spec's worked microworld rather than taking it from the PR — a claim used to justify *not* refusing something, which is exactly the kind that flips the correct behaviour if it is wrong.

## Receipts

Both at `3d2aa5d`, the PR head.

```
npm run gate
```
(exit 0 — 1057 passed | 7 skipped across lang/core/stores, 40 passed in cli; 1048 before this PR)

```
npm run e2e
e2e: green (pottery-studio installed from tarballs and played at 3d2aa5d)
```


ericeslinger:
**Non-blocking** — `sprout/lang/src/parse.ts:999`

The doc comment `` /** `visitors are Creature`, `visitors arrive at composing_room`. */ `` now sits directly above the new `private contains()` (line 1009), which has its own doc comment right below it — so this line is sandwiched between two doc comments for a different method, describes neither the block it precedes, and leaves `private visitors()` (line 1019) with no doc comment of its own. Cosmetic only; move it to sit directly above `visitors()`.

---

ericeslinger:
**Review**: no blocking findings, 1 non-blocking (misplaced doc comment, `parse.ts:999`). Gate at 0a56699: green, 1057 passed | 7 skipped (sprout), 40 passed (cli) — matches the PR's receipts exactly. Small, well-tested parser + world-resolution change; tried hard to find a swallowing bug in the new `contains()` reader and a surviving mutant in the idempotence logic and could not.

What I checked and did not find a problem:

1. **The `contains()` reader / swallowing risk (highest priority).** Ran the actual built parser (not hand-traced) against ~20 constructed inputs: `contains`/`contains actors` as the last member before `}`; both forms at EOF with the world never closed; `contains` immediately followed by a nested `enum` (the pre-existing #58/#59 "never closed" recovery path); a property `:actors 5` and a remembered `:remembers [actors: 5]` right after a bare `contains`; `contains actors actors` (repeated word — correctly refuses the second `actors` as an unknown member, and cleanly recovers to the closing brace); capitalized `contains Actors` (correctly not swallowed — kind-token mismatch, refused with "starts with a capital"); bare `actors` alone (correctly refused, it is not independently a member); a `:remembers` list with `contains` and `actors` as property names (both directions, with and without a preceding malformed entry); `contains` as a world's own name, and as a qualified kind's library segment (`world w: contains.Voice`); an enum option named `contains`. Every case matched the correct, non-swallowing behavior. I could not construct an input where `contains` on its own line swallows the next line's legitimate content, or where a legitimate `actors` occurrence is misread — `take('name','actors')` only ever consumes the single token immediately following the `contains` keyword, and there is no legitimate grammar position where a bare `actors` token can follow `contains` other than the capability itself (properties require a leading `:`, remembered-property names are read inside a wholly separate `:remembers` code path that never calls `contains()`).

2. **Whether this disturbs B12's `atDeclarationStart`/`atRecoveryStop` split or the #69 invariant.** It does not, structurally: both functions key off `this.readers` (the file-level `enum`/`message`/`world` map), never off the `worldMembers()` map that gained `contains`. Confirmed this isn't just a reading — the pre-existing test `#69 — a declaration's own word is an ordinary name...` already had a fixture (the `nameless` case in 'reports the second one even where its own word cannot say what it is') with a literal `contains` token inside a nameless `world: v.X { contains }` recovery block, predating this PR, and it still passes: that world fails before ever reaching the member loop (no name), and generic recovery skips straight over `contains` because it isn't in `this.readers`. `contains` remains fully usable as an ordinary name everywhere tested.

3. **The `contains actors ⇒ contains` invariant on `KindRef`.** Grepped every `KindRef` reference in the whole repo (not just `sprout/lang`): only `world.ts` (resolveWorld), `bindings.spec.ts`, and `check.spec.ts` construct one, and all three correctly derive `contains` from `contains || containsActors`. `check.ts`'s `container()` (unmodified by this PR) still reads plain `.contains`. Built a standalone `KindRef` fixture with `containsActors: true` and ran it through the real (unmodified) `typeOf`/`self.count` path — it type-checks to `integer` with no diagnostics, confirming the invariant holds end-to-end through code this PR doesn't touch, not just where it does.

4. **The three recorded spec gaps, and the "not a hole" claim.** Checked against the actual spec text, not the issue's paraphrase. Confirmed directly: `kind Place { contains actors ... }` and `object composing_room: sprout.Place` (which holds `cabinet`, `brass_key`, `apprentice`, `cat`) both appear in the spec's worked microworld, supporting "implies holding" as the only reading consistent with the spec's own example. Confirmed the "not a hole" claim is factually right: `kind World { contains ... }` is written explicitly in the spec's worked-microworld standard library (around line 1653), so "`sprout.World` declares `contains` itself" is true, not asserted. The *How members combine* table's single row for `contains, contains actors | idempotent` genuinely doesn't address single-author repetition, and *Places inside places* genuinely doesn't address a world becoming a place — both are real gaps, both recorded as the PR body describes. I did not find anything resolved silently outside these four.

5. **Mutation coverage.** Tried several 9th-mutant candidates against the `world.ts` accumulation: dropping the `||` in either line, in isolation or together, in every order of the four `contains`/`contains actors` combinations. Every variant I tried is caught by the existing `takes either of them twice, and both together, in silence` test, specifically because it exercises all four orderings, not just repetition. Also tried span-computation shortcuts in `contains()` (always spanning to the next token, using a zero-width span) — caught by the existing span assertion. Did not find a surviving 9th mutant.

Also checked B13's backlog line, which names a third clause ("exits only on places") the PR defers to B28 — legitimate, since no exit or grammar construct exists anywhere in the compiler yet to enforce that rule against; the PR says this explicitly rather than silently dropping it.

---

ericeslinger:
Fixed at `3d2aa5d`. You read it exactly right, including the cause: inserting `contains()` anchored on `private visitors()`, which pushed that method's one-line doc above the new one. `visitors()` has it back.

Two things in the review worth keeping rather than just acknowledging.

**The `#69` fixture that already covered finding 2.** I would have argued from the structure — `atDeclarationStart`/`atRecoveryStop` key off `this.readers`, not off `worldMembers()`, so a new *member* word cannot reach them. You went and found that a spec written two commits ago already had a literal `contains` token sitting inside a nameless-world recovery block, and checked it still passes. That is the difference between an argument and evidence, and it is the distinction that cost #68 three rounds.

**Checking the "not a hole" claim against the spec rather than the PR.** I asserted `sprout.World` declares `contains` itself; you went to the worked microworld and confirmed the line is there. A claim I make in a PR body to justify *not* refusing something is exactly the kind that should be checked against the source, since if it were wrong the right behaviour would be the opposite.

Receipts at `3d2aa5d`: gate exit 0 (1057 passed | 7 skipped sprout, 40 cli), e2e green. Body receipts updated. Merging.

---

PR #72, opened 2026-09-22
spec: sweep the holes Phases 0–1 recorded, as Eric decided them

ericeslinger:
## What and why

Eric answered every hole the working notes had recorded (the review page's "spec holes" tab, 44 comments, 2026-09-22). This PR writes those decisions into the spec and rewrites the notes' *Holes in the spec* section so it holds only what is still open. No code changes; each decision that changes code already built is listed below and gets its own issue.

Decisions carried into the spec:

- `sprout.World` is never written (on a world it is redundant, elsewhere wrong; both refuse).
- New *Lexical rules* under The compiler: `//` and `/* */` comments; escapes `\"`, `\\`, `\n`, `\{` in quoted text and passages (`{{` is gone); the colon rule; the reserved words, which may not name an option or a binding.
- Manifest: optional `namespace` falling back to `name`; `version` is semver; `author` is text; the `world` declaration repeats `name`.
- Blessed libraries are exempt from the file cap too.
- Which side of a verb knows the other is the author's choice.
- A default is always written; an option default may be qualified (`Ward.iron`) or, in a restatement, bare; `[]` is fine with a written element type.
- Enums: commas required, trailing comma allowed, host cap of 100 options, no reserved words as options.
- Lists may hold lists; two lists are not compared with `==`.
- Bindings table gains guard parameters, `$first`/`$last`/`$index`/`$count`, and the stock passages' `thing` and `candidates`.
- `count` on a list; a written precedence section.
- The claim that a `let` saves steps is gone; a `let` is charged as a statement.
- `chance(n)` is one in n; `random(n)` is 0 to n−1.
- Limits: unset means unbounded; a cap exceeded at load refuses unless the host has recorded an exception; the nesting cap is gone; spawns per hour is gone; live instances and pending wakes are stated as world-scoped rules rather than per-turn budgets.
- A missing arrival place is the host's refusal, not a passage.
- Diagnostic layout is the tool's. Levels are declared and requested, not derived.

Things I decided while writing that Eric should glance at:

- The reserved-word list is derived from the spec's own syntax. Say if any word should come off it.
- Eric wrote `Car[]` and `Ward [][]` in two comments; the spec keeps `[Car]`. Say if the postfix spelling is wanted.
- "Live instances per world" left the budgets table entirely on the reading that it is a host storage decision; "pending wakes: one per object" is stated as a language rule under Time. The question about riding it on a method/listener cap is recorded as open.
- The `world` declaration's name must equal the manifest's `name`; that sentence is mine.

B13 landed on main while this was open; its three containment holes are folded into the notes under "Recorded since the sweep, awaiting Eric".

Still open, per the notes: the colon's overloading (a proposal is written there), the roles design session (B23/B24/B26/B27 wait for it), the withheld-file hash, pending wakes, and three findings from the 2026-09-21 review.

Follow-up issues for code the spec now disagrees with: filed after this merges, one per bullet under "Decided, and different from what was built" in the notes.

## Receipts

- gate at `88b8688` (after merging main, which brought B13): `✓ corpus: every good archive passes, every bad one fails as expected`
- e2e at `2782ce4`: `e2e: green (pottery-studio installed from tarballs and played at 2782ce4)` (docs only; unchanged by the fix at `bf9dc16`)

Nothing fenced. Not a backlog item: a spec sweep Eric asked for directly; too small for an issue of its own by the house rule's "says in its PR why".


ericeslinger:
**Blocking** — `docs/design/sprout-design-spec.md:1473` (Lexical rules, reserved words)

The new reserved-word list omits words that are themselves literal syntax elsewhere in this same document, so the list does not actually match the syntax the spec uses.

- `as` and `for` are core body-declaration syntax — `as <role> for <verb>` (e.g. `as target for tag`, `as actor for take`, `as tool for unlock`) appears roughly 50 times through the spec, always as this bare keyword pair, not inside a passage `{…}`. Neither word is in the reserved list.
- `symbol` is used the same way `integer` is, as a literal value-role type: "A role may declare a value type instead of a kind: `symbol`, or `integer`." (`role topic: symbol`). `integer` (along with `boolean`/`string`/`object`) is captured under "the type names"; `symbol` is not, despite being written the same way in the same position.

Since the rule says a reserved word "may not name an enum's option or a binding," this gap means an author could legally write `role as: Kind` or `role for: Kind` or `role symbol: Lockable` — a role/binding whose name collides with the very keyword used to read it (`as <role> for <verb>`, `role x: symbol`), which the reserved-word rule exists to prevent. Add `as`, `for` and `symbol` to the list (or say why they're excluded).

ericeslinger:
**Review**: 1 blocking, 0 non-blocking. Gate at 2782ce4: green (lint, format, both builds, all suites, spec typechecks, corpus check all pass — matches the PR's receipt). Docs-only sweep of ~44 spec holes into the design spec and a rewritten *Holes in the spec* section; checked for residual `{{`, `chance(30)`, the removed nesting cap, "spawns per hour," and "decides for itself" (none found), cross-checked every bullet in the old Holes list against the new Still-open/Decided/Confirmed sections (nothing genuinely open appears dropped), and checked the reserved-word list against the spec's own syntax, where it is missing `as`, `for` and `symbol`.

ericeslinger:
Fixed in `bf9dc16`: `as`, `for` and `symbol` added to the reserved words. Gate at `bf9dc16`: `✓ corpus: every good archive passes, every bad one fails as expected`. e2e unchanged (docs only).

ericeslinger:
Merged main (B13, #70) into the branch at `88b8688`; the notes conflict was in the Holes section, resolved by keeping the sweep and adding B13's three containment holes under "Recorded since the sweep, awaiting Eric". Gate at `88b8688`: green, last line `✓ corpus: every good archive passes, every bad one fails as expected`.

ericeslinger:
Correction on the receipt above: the gate line I quoted at `88b8688` was read before that run had finished. Re-ran the gate at `9a8b798` (the squash of this PR on main) in a fresh worktree: exit 0, last line `✓ corpus: every good archive passes, every bad one fails as expected`.

PR #87, opened 2026-09-22
Wipe the board: delete the previous language, lay out the new one, strip history from comments, a stronger harness

### LINKED ISSUE #71: Wipe the board: delete the previous language, lay out the new one, strip history from comments, a stronger harness
Eric's call on 2026-09-22, after the review of Phases 0–1: the previous language's implementation is under a week old, runs nowhere, and is costing the rewrite a naming tax (`SproutType` for `ValueType`, `Gap` for `Absent`, `BundleOptions` for `CompileOptions`) and a flat directory that already has two parsers in it. It goes.

In one PR:

- **Delete** the previous language: `engine.ts`, `parser.ts`, `sprout.ts`, `sprout-lang.ts`, `definitions.ts`, `extensions.ts`, `grammar.ts`, `microworld.ts`, `sprout-skill.ts` and their specs in `sprout/lang`; `runtime.ts`, `turn.ts`, `scene.ts`, `testing.ts` in `sprout/core`; the two example worlds; `ext-media`; the old corpus; the CLI's `play`, `serve`, `skill` and the archive code they need. The store port, the memory store, the records and the conformance suites stay. The issues that were to replace each deleted thing (B22–B38, B45, B48, B49, B53) now build from nothing, which is what the spec asked for.
- **Lay out** `sprout/lang/src` by layer: `source/`, `syntax/`, `declare/`, `check/`, `bundle/`, `runtime/`. Each spec moves with its file.
- **Rename** what the old names had taken: `ValueType`, `Absent`, `CompileOptions`.
- **Strip history from comments**: a comment says what is true now; issue numbers, regressions and "before B12" go to git.
- **A stronger harness**: golden diagnostics over a corpus of new-language files, run by the gate; the "nothing vanishes" invariant from #66 as a shared helper; a `sprout check` that runs the new compiler so the CLI, the corpus and e2e exercise it from the first Phase 2 PR.
- **CLAUDE.md**: drop the legacy fence, add the layout, the comment rule and the harness, written so a smaller model does not regress them.

Closes the legacy fence in CLAUDE.md; supersedes the fencing plan for B49's e2e. Does not touch the spec.

ericeslinger:
## What and why

Eric's call on 2026-09-22 after the review of Phases 0–1: the previous language's implementation was under a week old, ran nowhere, and was costing the rewrite a naming tax and a flat directory with two parsers in it. This PR wipes it and shapes what is left for the work ahead. Closes #71.

**Deleted.** The previous language in `sprout/lang` (engine, command parser, old AST, old compiler, definitions, extensions, grammar, microworld, skill generator, their specs, the README-as-spec, the worked-example spec, fixtures); the old runtime in `sprout/core` (`runtime.ts`, `turn.ts`, `scene.ts`, `testing.ts`); the two example worlds and the `examples` package; `ext-media`; the old corpus; `SKILL.md`; the CLI's `play`, `serve`, `pack`, `skill`, zip, session, store and transcript. The issues that were to replace each of these (B22–B38, B45, B48, B49, B53) now build from nothing, as the spec asks.

**Kept.** The store port, its records, the memory store, the conformance suite and both store adapters. `records.ts` no longer imports the old language: `SproutState` is a local `z.record`, and the stored manifest is a record too. The records still describe the previous state model; B16 and B40 reshape them and the adapters follow. If you would rather the stores go too, say so and they go in a follow-up.

**Laid out.** `sprout/lang/src` by layer: `source/`, `syntax/`, `declare/`, `check/`, `bundle/`, `runtime/`. Each spec moved with its file. `KindRef` and `KindLookup` moved out of `bindings.ts`/`check.ts` into `declare/kinds.ts`, which removes the declare→check import. `index.ts` exports by layer.

**Renamed and simplified.** `SproutType` → `ValueType`, and the never-written `object` arm is gone from it (`HeldType` and `asHeldType` with it: the value type and the held type were the same thing once the arm went). `Gap` → `Absent`. `BundleOptions` → `CompileOptions`. Test-only parse entry points are still exported; the harness doc says they should not be, and that is a small follow-up.

**The manifest.** `bundle/manifest.ts` reads `sprout.json` with zod and reports at the key; `namespace` is added to `Manifest`, optional in the file and defaulting to `name`, and the world's own declarations compile under it (the spec sweep, #72).

**The CLI** is `init` and `check` on the new compiler. `check` prints the diagnostics page and a summary line, or JSON; `init` writes a manifest, a world and a README that pass `check`. A folder's vendored libraries have no on-disk layout yet; the notes' Holes should record that (one line, in the next docs PR; not done here because this PR does not touch the spec or notes).

**The harness.** `docs/design/sprout-test-harness.md` says what each layer is for. Implemented here: the corpus is golden pages — `corpus/good/*` pass, `corpus/bad/*` fail with exactly `expected.txt` — run by `npm run check`; e2e installs the tarballs and runs `init` and `check` from the installed CLI. Still to build, and listed there: golden transcripts (B49), the author-facing test format (B52), replay determinism, and a shared generated-input helper.

**Comments.** Every file under `sprout/lang/src`, `sprout/core/src` and the store packages had history stripped from its comments: issue numbers, regression stories, "the previous language", citations of the superseded split proposal. Three subagents did it, comments and test titles only, verified per file against reversed replacements; no assertion or code line changed. Comment lines went from 20–50% of each file to roughly half that.

**CLAUDE.md** drops the legacy fence and adds *Where code goes* (the layer table and its import rule, the 800-line split rule, functions over a context), *Comments* (present tense only, with the concrete forbidden forms), *The harness*, a "stop after two rounds on one finding" rule, and the roles hold. `.claude/agents/pr-review.md` checks the same things in place of the fence. READMEs rewritten to match.

**B13** (#70) landed on `main` while this was open and is merged in: its hunks applied to the moved files, and `containsActors` lives in `declare/kinds.ts`.

## Receipts

- gate at `3f6a3ef`: `✓ corpus: every good world passes, every bad one fails as expected`
- e2e at `3f6a3ef`: `e2e: green (init and check from the installed CLI at 3f6a3ef)`

## For Eric

- The stores: kept and decoupled, or delete too? (See *Kept*.)
- A cloud session is working the backlog against the old layout (B13 merged mid-wipe). Anything it has in flight will need rebasing onto this; B14 in particular touches `world.ts` and `parse.ts`.


ericeslinger:
**Blocking** — `sprout/lang/src/bundle/manifest.ts:56-61` (no colocated spec)

`parseManifest`'s schema-refusal branch (the `ManifestSchema.safeParse` failure path, and `manifestKeySpan`) has no spec anywhere — not `manifest.spec.ts` (there isn't one), and not transitively: `world.spec.ts` only exercises the JSON-parse-failure path and the happy path. CLAUDE.md's *Where code goes* requires "a spec file beside every source file … that exercises it directly; coverage through another file's spec does not count," and `manifest.ts` is new in this PR.

The gap is not cosmetic. Feeding an invalid manifest through the built CLI produces a broken sentence:

```
$ sprout check .   # sprout.json has "level": "one"
sprout.json:6:3  The manifest's level is Invalid input: expected number, received string.
                 A manifest is a JSON object with a name, a version, an author, a license, a level and a list of files.
```

and, for a missing required key:

```
sprout.json:1:1  The manifest's license is Invalid input: expected string, received undefined.
```

`${issue.message}` is zod's raw English ("Invalid input: expected number, received string"), spliced mid-sentence with a capital letter after "is". This doesn't read as a sentence and doesn't say what to write instead — it fails the spec's Diagnostics section ("Messages are written for someone who is not a programmer, and say what to write instead") and CLAUDE.md's invariant that a diagnostic tells a non-programmer what to write. A colocated spec asserting the exact message (as the corpus does for every other refusal) would have caught this.

Fix: write the manifest's own message per zod issue code (or at least drop the raw `issue.message` and name the expected shape, e.g. "`level` is a whole number." / "`license` is missing, and every manifest has one."), and add `manifest.spec.ts` covering JSON-parse failure, a missing key, a wrong-shaped key, and the key-span lookup.

ericeslinger:
**Non-blocking** — `sprout/lang/src/declare/kinds.ts` (no colocated spec)

New file, no `kinds.spec.ts`. `kindName` and `composesKind` are exercised directly by name from `check/bindings.spec.ts` (lines ~160-178) and `declare/world.spec.ts`, so the behaviour is verified, but per CLAUDE.md ("coverage through another file's spec does not count") the file still wants its own colocated spec. Worth adding a small `kinds.spec.ts` that owns these tests directly, since `bindings.spec.ts` and `world.spec.ts` importing from `../declare/kinds.js` to test kinds.ts's own exports reads like exactly the pattern the rule is written against.

ericeslinger:
**Non-blocking** — comment style across most of the newly laid-out files (e.g. `sprout/lang/src/declare/kinds.ts:1-2`, `sprout/lang/src/check/bindings.ts:17,25-27`, `sprout/lang/src/check/check.ts:27-30`, `sprout/lang/src/declare/world.ts:9-11`, `sprout/lang/src/runtime/budget.ts:7,16`, `cli/src/cli.ts:7`)

This PR's own new CLAUDE.md says a comment "never says how the code got here" and lists, unconditionally, "No issue or PR numbers." Dozens of the moved/rewritten file headers and doc comments still cite backlog items by number ("B19 composes kinds and B21 namespaces them", "verbs and their roles are B23's", "B14 resolves `visitors arrive at`, B32 reads the pass rules, and B42 handles arrival", "`play` and `serve` return when the runtime does (B34, B37)"). These are forward pointers to future work rather than history, so they read differently from the "before B12" case the rule names, but the rule as written doesn't carve out an exception for them, and a PR that writes the rule and then violates it forty times in the same diff is worth a second look — either scrub these (git/the backlog already says what's next) or narrow the rule in CLAUDE.md to say forward references are fine and only history is banned.

ericeslinger:
**Non-blocking** — `corpus/bad/enum-twice/kinds.sprout`

The input has two problems (`Ward` lists `oak` twice, and `Ward` is declared twice in one file), but `expected.txt` pins only the first. That's correct as written — `compile.ts` drops a file's declarations from every table once its shape check refuses, so the second `enum Ward` block never reaches `EnumTable.add`'s "declares two enums called X" refusal — but it makes the corpus entry read as if it's pinning the duplicate-declaration diagnostic when it isn't; that path has no `corpus/bad` coverage at all (only a unit test in `enums.spec.ts`). Either drop the second `enum Ward` block (it's dead weight for what this test title suggests) or add a separate corpus world for the duplicate-declaration refusal.

ericeslinger:
**Review**: 1 blocking, 3 non-blocking. Gate at `7ec2b0e`: green (lint, prettier, both builds, 880 tests, spec typechecks, corpus — matches the PR's receipt). `npm run e2e` at the same sha: green (init and check from the installed tarballs). A large mechanical PR — deletion, layout, renames (`SproutType`→`ValueType` with the object arm dropped, `Gap`→`Absent`, `BundleOptions`→`CompileOptions`), comment-stripping, a new corpus and harness — and it holds up under scrutiny: spot-checked the largest-diff files (`bindings.ts`, `check.ts`, `world.ts`, `types.ts`, `compile.ts`, `parse.ts`, and every touched `core`/`store-sql`/`store-document` file) with `diff -w` against `origin/main` and found no behaviour change beyond the stated renames, the new `namespace` field (per the spec's manifest table and #72), and the described `records.ts` decoupling; the layer-import table (`declare/` never imports `check/` or `bundle/`, `syntax/`→`bundle/limits` allowed) is obeyed everywhere checked; `sprout/package.json`'s `exports`/`files` match what actually builds and packs. The one blocking finding is a real gap: `bundle/manifest.ts` has no colocated spec and its schema-refusal path produces a broken, non-professional diagnostic (verified by hand against the built CLI) — the kind of bug a colocated spec exists to catch.

ericeslinger:
All four answered in `3f6a3ef`; gate and e2e re-run at that head (receipts in the body).

1. **Blocking, manifest.ts** — fixed. The refusal now says what is wrong in plain words ("The manifest's version is not a string.", "The manifest has no level.") and the remedy shows the field to write ("Write \"level\": 1."). `bundle/manifest.spec.ts` covers the good path, the namespace default, not-JSON, a wrong-shaped field at its key, missing fields, and the remedy text.
2. **kinds.ts spec** — added `declare/kinds.spec.ts`; the kind-matching tests moved there from `bindings.spec.ts`, plus a lookup case.
3. **Comment rule vs. backlog numbers** — the rule now says it: `#nn` issue and PR numbers are out; a backlog item named as the owner of work not yet done ("B29 reads passage bodies") stays until it lands. That is the reading the cleanup applied.
4. **corpus/bad/enum-twice** — split: `enum-twice` is now two files each declaring `Ward`, so the table's refusal is the one that fires; `option-twice` holds the duplicate-option case.

Issue #82, opened 2026-09-22
`sprout.World` is written on every world and refused anywhere else


ericeslinger:
Eric restated this on 2026-09-22, reversing the earlier reading: `sprout.World` is **required** on the root world, written explicitly — `world printers_shop: sprout.World { … }` — and **refused** on anything but a world. A world composes like a kind, so it may compose more beside it: `world w: sprout.World, victorian.Voice { … }`. Level 1 is explicit wherever it could have defaulted, because a later level can relax a requirement and never add one.

Today `resolveWorld` composes `sprout.World` implicitly and accepts an explicit one as redundant. Change it to refuse a world whose composition list lacks `sprout.World` (naming what to write), keep it first in the composition order whether or not it was written first, and have the kind resolver (B19) refuse `sprout.World` in any kind's or object's composition. Update `corpus/good/declarations/world.sprout`, `sprout init`'s world, and every spec fixture that writes a bare `world w { … }`; add a `corpus/bad` world for the missing case.

Spec: The world model (first section), The compiler › What it refuses, Language levels.

ericeslinger:
Clarified with Eric on 2026-09-22 before starting: the composition must be the literal `sprout.World`; an unqualified `World` does not satisfy it, and a later level may relax that. Kinds are not resolved until B19, so the check reads what was written, in the shape tier. The other half, refusing `sprout.World` in a kind's or an object's composition, has nothing to check until B19 parses kinds; it is recorded on #19 and does not hold this issue open.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #88, opened 2026-09-22
spec: sprout.World is written on every world, and refused anywhere else

ericeslinger:
## What and why

Eric restated the `sprout.World` rule on 2026-09-22, reversing the reading the sweep (#72) recorded: it is **required** on the root world, written explicitly, and **refused** on anything but a world. A world composes like a kind, so it may compose more beside it: `world printers_shop: sprout.World, victorian.Voice { … }`.

Changes, all docs:

- The world model's opening paragraph states the rule; the three world examples in the spec now write `: sprout.World`.
- *What it refuses* gains the two refusals (a world without it; anything else with it).
- *Language levels* states the principle behind it: level 1 is explicit wherever it could have defaulted, because a later level can relax a requirement and never add one to worlds already accepted.
- The working notes' "Decided, and different from what was built" bullet is corrected, with the reversal noted.
- Follow-up issue #82 is retitled and rewritten to the new rule (it had said the opposite).

Not a backlog item: a spec decision Eric made directly; too small for an issue of its own.

## Receipts

- gate at `05f54f5`: `✓ corpus: every good world passes, every bad one fails as expected`
- e2e at `05f54f5`: `e2e: green (init and check from the installed CLI at 05f54f5)`


ericeslinger:
**Review**: no blocking findings. Gate at 05f54f5: green (865 sprout + 26 cli tests passed, typecheck:spec clean, corpus check green — matches the PR's receipts).

Docs-only change, scoped to the two files listed. Checked: all three world declarations in the spec now write `: sprout.World` (grep for `sprout.World` at the PR head turns up no remaining "implicit"/"never written" language); the new refusal bullet in *What it refuses* matches the prose in *The world model*; the *Language levels* addition cites already-established facts (`a default is always written`, type inferred from a literal) as illustration rather than introducing new behavior; the working notes' "Decided, and different from what was built" bullet states the same rule as the spec, and the nearby "Not a hole" bullet is updated to note it's now explicit. Issue #82 was retitled/rewritten to match, consistent with the PR body's claim. `corpus/good/declarations/world.sprout` still omits `sprout.World` and the checker doesn't yet enforce the rule — expected, since that's #82's job and this PR is spec/notes only.

PR #89, opened 2026-09-22
spec: tools, optional tools and `bound`

ericeslinger:
## What and why

The first of the roles questions, settled with Eric on 2026-09-22 in conversation. Docs only.

- **Tools.** A verb's first role is the **target** and every other role is a **tool**, whether a thing or a value; the word is called out as a convenience. The `role` keyword is unchanged (Eric did not ask for a syntax change; flagging that I kept it).
- **Verbs with no tools, or no roles**, are stated explicitly rather than implied.
- **Optional tools** (new subsection under Verbs). A tool some phrase leaves out is optional; a tool every phrase fills is not. An optional tool is read only inside `if (bound x)`, which narrows it the way `is()` does; reading it elsewhere is a refusal that names the phrase and says what to write. No value stands for an unbound tool. A phrase-less verb declares `role x optional`; `act` may omit an optional tool and may not omit one that is not. A set role that a phrase leaves out is still the empty set, never optional.
- **A `symbol` tool with no `from`** is never bound for that role-player and may not be read at all, `bound` included.
- **Every value tool is optional** (second commit, `13b9517`, Eric's second decision): a symbol among the `from` options or an integer inside the `from` range binds; anything else the visitor supplies still matches the phrase and arrives unbound (*ask the guard about potatoes* runs the `else` branch, carrying nothing of what was typed); no `from` means never bound, for integers as for symbols. That closes the first two of the three roles questions; the notes say only `many` on a value tool remains.
- Bindings table, checker table, refusals and reserved words (`bound`, `optional`) updated to match. The notes' *Roles* entry now says what is settled and what the two remaining questions are.

Not a backlog item: a spec decision Eric made directly.

## Receipts

- gate at `b8a458c`: `✓ corpus: every good world passes, every bad one fails as expected`
- e2e at `b8a458c`: `e2e: green (init and check from the installed CLI at b8a458c)`


ericeslinger:
Head moved to `13b9517` with Eric's second decision (every value tool is optional; a value outside its options or with no `from` arrives unbound). Gate at `13b9517`: green, `✓ corpus: every good world passes, every bad one fails as expected`; e2e at `13b9517`: green. The review in progress was pinned to `47a4172`; the delta is the Value roles and Optional tools paragraphs and the notes' Roles entry.

ericeslinger:
**Blocking** — `docs/design/sprout-design-spec.md:811-825` and `:2016-2021`

The second commit (`13b9517`) makes every value tool optional and says reading it "arrives with `topic` unbound" whenever the typed value isn't among the offered options — and the refusals list this same PR writes says "An optional tool read outside `if (bound x)`" is a compile error. Neither existing code example that reads a `symbol` tool was updated to match.

The `Guard`/`ask` example under Value roles reads `topic` directly in `do { if (topic == :toll) { … } else { … } }` with no `if (bound topic)` anywhere — the prose right below it even says "the guard's `else` branch is where 'he has never heard of potatoes' is written," but the code's `else` still just says `bridge_speech`, and the read itself is now outside the required guard. The worked microworld's `apprentice.as target for ask` (line ~2016) has the identical problem: `if (topic == :the_press) … else if (topic == :the_cat) …`, no `bound` anywhere, and no branch at all for an unbound topic (so *ask the apprentice about the loft* would also violate the silence invariant, on top of not compiling).

As written, the spec's own worked microworld — the document's canonical example, "written to exercise as much of this document as a small world honestly can" — would not compile under the rule this PR just added. Fix: wrap both reads in `if (bound topic) { … } else { … }`, with the `else` covering the unbound/off-topic case per the new prose.

ericeslinger:
**Non-blocking** — `docs/design/sprout-design-spec.md:1548`

The refusals bullet reads: "`bound` on a tool that is not optional, or on a `symbol` tool with no `from`." The second commit's own prose (line 837) draws an explicit parallel for integers — "no `from` at all[…] leaves the tool unbound, exactly as for a symbol" — and the preceding paragraph says a `symbol` with no `from` "may not [be read], not even under `bound`." If that's meant to hold for an `integer` tool with no `from` too, the refusals list should say "a `symbol` or `integer` tool with no `from`" rather than naming only `symbol`. Not blocking since no code implements this yet, but worth tightening so the table matches the prose's own stated symmetry.

ericeslinger:
**Review**: 1 blocking, 1 non-blocking. Gate at `13b9517` (current head — note the PR body's receipts were updated to this sha mid-review, after I'd started against the first commit `47a4172`): green, re-run in a worktree (35/36 suites, 1 skipped as before; corpus and typecheck:spec all pass). Docs-only change; e2e not re-run (no touch to cli/, sprout/core, packaging or corpus, and the receipt is at the current head).

Checked: the target/tool split, "no tools"/"no roles" verbs, the Optional tools subsection, `bound`/`optional` reserved words, the bindings and checker tables, and the two roles decisions against the worked microworld's own verbs (`unlock`, `work`, `ask`). `unlock`'s `tool` and `work`'s set `tools` are consistent with the new rules (every phrase in the microworld's `unlock` fills `tool`, so it's rightly not optional there; the set role is rightly never optional). The `ask`/`topic` value-role examples are not consistent with the second commit's "every value tool is optional" rule — see the blocking comment.

ericeslinger:
Both answered in `b8a458c`. **Blocking:** the guard's and the apprentice's `ask` bodies now read `topic` inside `if (bound topic) { … } else { say … }`, so the worked microworld compiles under the rule and an off-topic question is answered rather than silent. **Non-blocking:** the refusal now names a `symbol` or `integer` tool with no `from`. Gate at `b8a458c`: green, `✓ corpus: every good world passes, every bad one fails as expected`; e2e at `b8a458c`: green.

PR #92, opened 2026-09-22
spec: the decisions behind the alignment issues, as Eric answered them

ericeslinger:
## What

Docs only, no code: the spec, the backlog and the working notes brought to what Eric decided on 2026-09-22 when the alignment issues #73 to #86 were read back to him, before any of them is implemented. Each issue carries a comment saying what was clarified.

This PR edits the spec, which a PR normally does not; it does so because these are Eric's own decisions being written down, and it is assigned to him to accept or amend rather than merged by the agent.

**Spec**
- The world model, What it refuses: `sprout.World` is the literal spelling; an unqualified `World` does not stand for it (#82).
- The manifest, What it refuses, the absent table: a bundle has exactly one `world` declaration, named as the manifest's `name`; none, two, or another name is refused at publish, and at load the world admits no one, as for a missing arrival place (#83).
- Declaring a property: a default option may name its enum's library, `:ward sprout.Ward.iron` (#77).
- Lists: inside a list of lists, `add`, `remove` and `includes` take two elements to be the same when they hold the same elements in the same order; never an `==` (#78, #79).
- What the compiler checks: a symbol literal is checked on either side of `==` (#86, the settled half).
- Lexical rules: an unclosed `/* … */` is refused at its opening; block comments do not nest (#73).
- Limits: nesting has no cap; the compiler's recursion bound is its own and not recorded in the bundle (#80).

**Backlog rows** B02, B12, B18 and B29 no longer say live instances as a budget, `sprout.World` implicit, spawns per hour, or `{{`.

**Working notes, Holes in the spec**: the 2026-09-21 review bullet now records the symbol-literal side as settled and keeps the integer literal in a comparison open; new lines for the block comment, a library's `version` as semver (open), nested-list sameness, and the one-world rule; the decided-list bullets sharpened to match.

## Still open for Eric

- #86, integer half: whether `self.get(:wear) == 12` on a 0-to-9 property is refused as the compile-time literal case, or accepted because the spec's table refuses out-of-range literals only for `set` and `remember`.
- #75: whether a library's `version` is semver as the world's is.

Gate at 1bbf377: `✓ corpus: every good world passes, every bad one fails as expected`
e2e: not run; the diff touches only `docs/design/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UBMVBu1sWCjzBKX2aGmKeX

---
_Generated by [Claude Code](https://claude.ai/code/session_01UBMVBu1sWCjzBKX2aGmKeX)_

PR #93, opened 2026-09-22
Lexer: \{ is a fourth escape in quoted text

### LINKED ISSUE #74: Escapes: `\{` in quoted text, and the same escapes in a passage; `{{` is gone
The lexer takes `\"`, `\\` and `\n`; the spec now adds `\{` and says a passage takes the same escapes, with `\{` as its literal brace. B29 reads passages; the lexer's half can land first.

Decided by Eric on 2026-09-22 (the review page's spec holes tab) and written into the spec by #72. The code was built before the decision; this brings it in line. Spec section: The compiler › Lexical rules; Prose › Slots.

ericeslinger:
## What

Quoted text takes `\{` as a fourth escape, yielding a literal `{`, per the spec's The compiler › Lexical rules. The refusal for an unknown escape names all four. An unescaped `{` inside quotes stays an ordinary character; only a passage gives it meaning, and passages are B29's, so the passage half of the rule (the same escapes, `\{` as the literal brace, `{{` gone) is not touched here beyond the backlog row #92 already corrected.

## Spec gaps

None.

## Evidence

- `corpus/good/declarations/world.sprout` gains a string default written with `\{`.
- `sprout/lang/src/syntax/lexer.spec.ts` covers the four escapes and an unescaped `{` inside quotes.

Closes #74

Gate at c67dd14: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at c67dd14: `e2e: green (init and check from the installed CLI at c67dd14)`

---


ericeslinger:
**Review**: no blocking findings. Gate at c67dd14: green (e2e also green). Small, spec-faithful lexer change adding the fourth `\{` escape to quoted text, with matching corpus/good and lexer.spec coverage.

_(Posted on behalf of the `pr-review` agent, which cannot post here itself in this session. It pinned itself to refs/pull/93/head at c67dd14 and re-ran the gate and e2e in its own worktree. It also noted, not as a finding, that the unknown-escape refusal has no `corpus/bad` world, which predates this PR.)_

---

PR #95, opened 2026-09-22
Enum options: a trailing comma, the host's cap, and no reserved words

### LINKED ISSUE #76: Enum options: allow a trailing comma, cap at 100 per host, refuse reserved words
The parser refuses a trailing comma after the last option; there is no options cap; an option may be any lower-case word. The spec now allows the trailing comma, adds `options per enum | 100` to the static caps (a host default in `limits.ts`), and forbids the reserved words as options.

Decided by Eric on 2026-09-22 (the review page's spec holes tab) and written into the spec by #72. The code was built before the decision; this brings it in line. Spec section: Properties › Enums; Limits › Static caps; The compiler › Lexical rules.

ericeslinger:
## What

Three rules from the spec's Properties › Enums, Limits › Static caps and The compiler › Lexical rules:

- A comma after the last option is allowed: `enum Ward { oak, silver, }` has two options. `{ , }`, `{ oak,, silver }` and a missing comma keep their refusals.
- `options per enum | 100` is a static cap in `limits.ts` (`optionsPerEnum`, in the limit table with scope `enum`). `checkEnumDeclaration` takes the host's figure and refuses once, at the first option past the cap, naming the count and the cap. The figure is passed in as a number because `declare/` may not import `bundle/limits`; `checkShape` resolves the host's caps once and passes it.
- The reserved words live in `syntax/reserved.ts`, in the spec's order; its spec reads the sentence out of the spec file, so the two cannot drift. A reserved word in an option's place is still read as a word, so the enum keeps its other options and the declarations after it survive, and is then refused at the word. A missing comma before such a word is still named, at its gap, before the word is refused. A reserved word as a `let` name is refused too, per Lexical rules' "None may name an enum's option or a binding" (Eric, 2026-09-22, on the issue). Property names are not bindings and are left alone.

The parser's comments that said nothing reserves an option's name now say what is true: the lexer reads every reserved word as a plain name, and the reader that gives a name to something is what refuses it.

## Spec gaps

None new. Which existing tests changed meaning: a reserved word at the end of an unclosed enum now draws two refusals (the word, and the missing brace), both true of what is written.

## Evidence

- `corpus/good/declarations/kinds.sprout` gives one enum a trailing comma; `corpus/bad/option-reserved` and `corpus/bad/options-too-many` pin the two new refusals' pages.
- `parse.spec.ts`, `enums.spec.ts`, `limits.spec.ts`, `reserved.spec.ts` cover the trailing comma, the cap at exactly the option past it (with a small host cap), the reserved word as an option and as a `let` name, that the declarations after a reserved option are not lost, and, over generated enum bodies mixing ordinary and reserved words, that every missing comma is reported exactly once.

Closes #76

Gate at b71f746: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at b71f746: `e2e: green (init and check from the installed CLI at b71f746)`


ericeslinger:
**Blocking** — `sprout/lang/src/syntax/parse.ts:487-508`

A missing-comma diagnostic between the previous option and a reserved word is silently dropped whenever the reserved word is the last thing before `}`, and silently overwritten (also dropped) when another option follows it without a comma.

Reproduced directly against the PR head's `parseDeclarations`:
- `enum Ward { oak silver bronze }` (no reserved words, two missing commas) → correctly reports two `` `Ward` needs a comma between its options. `` diagnostics.
- `enum Ward { oak message }` (missing comma right before the reserved word `message`, which is also the last option) → reports only `` `message` is a word of the language, so it cannot be an option of `Ward`. `` — the missing-comma refusal never fires, even though the comma really is missing.
- `enum Ward { oak message silver }` → reports the reserved-word refusal plus exactly one missing-comma refusal (for the message/silver gap); the oak/message gap's `missingComma` marker is silently overwritten and never reported.

Cause: the `if (isReserved(word.text)) { … refused = true; }` branch never inspects or clears the pending `missingComma`, unlike the `else` branch, which checks it before pushing the option. Because `missingComma` is later reassigned only when `this.separator('}')` itself returns `'missing'`, a gap that occurred immediately before a reserved word is lost as soon as the reserved-word branch runs.

This is a regression introduced by this PR: before it, `message`/`enum`/etc. were ordinary option words, so the missing-comma path always ran for them. The generated-input invariant in `parse.spec.ts` only checks "never throws" and "never says one thing twice in one place", so it does not catch a diagnostic that never appears.

Fix direction: check and report `missingComma` in the reserved-word branch too, the same way the `else` branch does, so a real missing comma is never lost because the word after it happens to be reserved.

---

ericeslinger:
**Review**: 1 blocking, 0 non-blocking. Gate at 5479616: green. e2e: green. The reserved-words table (`syntax/reserved.ts`), its spec-file-derived test, the `optionsPerEnum` cap wiring through `bundle/limits.ts` → `bundle/compile.ts` → `declare/enums.ts` (kept out of `declare/`'s import restrictions, taking a plain `number`), the trailing-comma grammar change, and the `let`-name reserved-word refusal all check out against the spec and the layering rules; `checkShape`'s `caps ?? DEFAULT_LIMITS.caps` preserves every prior caller's behavior; the corpus fixtures are accurate and read correctly; no stale "nothing reserves an option's name" comment remains in a context the PR's change touches. The one blocking issue is a diagnostics-completeness bug in the new reserved-word branch of `enumDeclaration()`'s option loop: it silently drops a real missing-comma refusal when a reserved word sits where a comma was needed.

_(Posted on behalf of the `pr-review` agent.)_

---

ericeslinger:
Fixed in b71f746. The missing-comma report is hoisted above the reserved-word branch so it runs for any word that read, at the gap it marks and before the word is refused; the remedy lists the options read so far and `…`, leaving out a word the author has to replace anyway. Tests for the three cases (`{ oak message }`, `{ oak message silver }`, `{ oak message, silver }`) assert the messages in order and each comma refusal's line and column, and a generated-input invariant over enum bodies mixing ordinary and reserved words asserts one missing-comma refusal per spaced gap; reintroducing the bug fails it. Receipts in the body are at b71f746.

---

ericeslinger:
**Review**: 0 blocking, 0 non-blocking. Gate at b71f746: green. The fix hoists the missing-comma report above the reserved-word branch as claimed, is confirmed against all three named repro cases and a from-scratch edge case (`{ oak message default }`, two reserved words each after a missing comma), and the new invariant plus example tests were verified, by running them against the pre-fix parser, to fail on the reintroduced bug, including a subtler pre-fix misbehaviour (a stale, wrongly-anchored comma diagnostic) beyond what the original blocking finding described. Diff is scoped to exactly the claimed two files. The blocking finding above is resolved.

---

PR #96, opened 2026-09-22
A default option may be written qualified: `:ward Ward.iron`

### LINKED ISSUE #77: A default option may be written qualified (`Ward.iron`), and bare in a restatement
`:ward Ward default iron` is the only spelling read. The spec now also reads `:ward Ward.iron`, and reads a bare `:ward iron` where the property restates one already declared (B19's path) and the type is known.

Decided by Eric on 2026-09-22 (the review page's spec holes tab) and written into the spec by #72. The code was built before the decision; this brings it in line. Spec section: Properties › Declaring a property.

ericeslinger:
## What

A property may name its enum and the option it starts at as one, `:ward Ward.iron`, or with the enum's library, `:ward sprout.Ward.iron`, per the spec's Properties › Declaring a property. The parser fills the declaration's `type` and `default` from that one form, so nothing below `syntax/` sees a new shape; `resolveProperty` and the option check are untouched. `min`/`max` after it are read and then refused as for any symbol.

Refused, each at its token with a remedy: `default` after the qualified form (`:ward` says its default twice); a dot after a type that has no options (`:n integer.3`, `:opens [Ward].oak`); nothing, or a capitalised word, after the dot. The remedy for a bare option with no enum now offers both spellings, the shorter first.

The bare `:ward iron` in a restatement is B19's and is not built here; a bare option with no written type is refused as before.

## Spec gaps

None. Whether a world may qualify its own enum with its own namespace is not something the spec says, so the corpus does not use it.

## Evidence

- `corpus/good/declarations/world.sprout` writes `:ward Ward.iron`; `corpus/bad/default-twice`, `corpus/bad/default-dot-misplaced` and `corpus/bad/default-option-capitalised` pin the three refusals' pages (one each, because the world reader reports only the first bad member: #97).
- `parse.spec.ts`: both spellings parse with the option at its own line and column, including inside `:remembers`; each refusal at its token; the "a defect never loses a well-formed neighbour" table gains the four new defects. `properties.spec.ts`: the qualified spelling resolves to the same property as `Ward default iron`; `Ward.slver` is refused naming the options; `sprout.Ward.iron` resolves through the library.

Closes #77

Gate at 9104e07: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at 9104e07: `e2e: green (init and check from the installed CLI at 9104e07)`


ericeslinger:
**Non-blocking** — `sprout/lang/src/syntax/parse.ts:789-841`

Two of the three new refusal messages this PR introduces have no `corpus/bad` world pinning their exact words — only colocated unit coverage in parse.spec.ts. `corpus/bad/default-twice` pins "says its default twice", but "writes a dot after a type that has no options" and "`X.` cannot name …" are untested at the corpus/golden layer. The words read correctly, so this is not blocking, but per CLAUDE.md's harness section a new refusal worth pinning gets a `corpus/bad` world.

**Non-blocking** — `sprout/lang/src/declare/types.ts:154-159`

`typeOfLiteral`'s remedy for a bare option-literal default (`:state oak` with no type at all) still says only "Write the enum before the default, as in `:state Drying default oak`." Now that `:state Drying.oak` is an equally legal, shorter fix for exactly this mistake, the diagnostic teaches an author the longer of two correct spellings.

**Review**: 0 blocking, 2 non-blocking. Gate at 9d61877: green. E2e at 9d61877: green. Small, well-scoped parser change (qualified enum-default spelling reusing the existing property AST fields); traced the parsing and recovery paths for `atType()`, the dot branch, `:remembers`, and the extended defect table by hand and found the invariant and shared-code reuse hold.

_(Posted on behalf of the `pr-review` agent, which cannot post here itself in this session. Head has since moved to b3e825f by merging main; the diff is unchanged.)_

---

ericeslinger:
Both non-blocking findings are addressed, at 9104e07. `corpus/bad/default-dot-misplaced` and `corpus/bad/default-option-capitalised` pin the two refusals' pages, one each because the world reader reports only the first bad member in a body (filed as #97, pre-existing). `typeOfLiteral`'s remedy now reads "Write the enum with it, as in `:state Drying.iron` or `:state Drying default iron`." One intermediate commit (67e88a4) went out with a red test in `types.spec.ts` because my gate check read the wrong exit status; 9104e07 corrects it and the receipts in the body are at that head.

---

Issue #79, opened 2026-09-22
Two lists are not compared with `==`; drop the runtime's `equals`


ericeslinger:
`sameType` lets two lists compare and `SproutList.equals` answers it by order. The spec now refuses `==` on lists; set and ordered equality are a later level's.

Decided by Eric on 2026-09-22 (the review page's spec holes tab) and written into the spec by #72. The code was built before the decision; this brings it in line. Spec section: Properties › Lists, What the compiler checks.

PR #98, opened 2026-09-22
A list's element type may itself be a list

### LINKED ISSUE #78: A list's element type may itself be a list
`resolveType` refuses `[[Ward]]`. The spec now allows it, every element of the one type.

Decided by Eric on 2026-09-22 (the review page's spec holes tab) and written into the spec by #72. The code was built before the decision; this brings it in line. Spec section: Properties › Lists.

ericeslinger:
## What

`resolveType` accepts `[[Ward]]` and deeper, per the spec's Properties › Lists. At run time `SproutList`'s `Element` includes a list, and the no-duplicates rule inside a list of lists is kept by a private sameness: two elements are the same when they hold the same element type, as many elements, and elements that are the same in order. `of`, `includes`, `add` and `remove` use it. It is never an `==` an author can write, which is what Eric decided on 2026-09-22 (on the issue and in the spec's Lists). The list cap applies to each list separately, inner ones included.

`check.ts` needed nothing: `sameType` already recurses, and the `add`/`remove`/`includes` rules are element-type-agnostic; tests pin that `self.add(:grid, self.get(:row))`, `remove` and `includes` type-check with `:grid [[Ward]]` and `:row [Ward]`, and that a symbol where a `[Ward]` is wanted is refused with the existing words. Compile-time duplicate detection of a nested literal already worked through `elementKey`.

`SproutList.equals` still compares elements by identity and is therefore wrong for nested lists; #79, next in line, deletes it as the spec's "two lists are not compared with `==`" requires. It has no caller but its own spec.

## Spec gaps

None new; the sameness rule is already in the spec and the working notes.

## Evidence

- `corpus/good/declarations/world.sprout` declares `:grid [[Ward]] default [[iron], [brass, iron]]`. No `corpus/bad` page changed; the old "A list holds values, not other lists" refusal had none.
- `types.spec.ts`: nested and triply nested types resolve; `[[oak], [oak]]` is refused as a duplicate at the second element; `[[oak], oak]` at `oak` as the wrong shape. `lists.spec.ts`: `of` drops a later same-order duplicate and keeps a different-order one; `includes`/`add`/`remove` with a nested element; `toString`; the cap on an inner list. `check.spec.ts` as above.

Closes #78

Gate at a3386fd: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at a3386fd: `e2e: green (init and check from the installed CLI at a3386fd)`

---


ericeslinger:
**Review**: no blocking findings. Gate at a3386fd: green. e2e at a3386fd: green. Small, well-scoped PR: nested list types resolve correctly, dedup/`same()` for lists-of-lists is order- and type-sensitive per spec (checked against inner lists differing only in declared element type, and empty inner lists), caps apply per nesting level, `Element`'s widening leaks no `any`, `[[oak], [oak]]` is refused at the second element, and the one known-wrong path (`equals()`) has no caller but its own spec and is deleted by #79, already branched off this PR.

---

PR #99, opened 2026-09-22
Two lists are not compared with `==`; the runtime's `equals` goes

### LINKED ISSUE #79: Two lists are not compared with `==`; drop the runtime's `equals`

ericeslinger:
## What

The checker refuses `==` and `!=` between two lists of the same type, at the left operand, per the spec's What the compiler checks ("same type, and not a list") and Lists ("Two lists are not compared with `==`; whether equality is by set or by order is a later level's to add"). The remedy points at what a list does answer: `includes` and `count`. A list against anything else falls through to the existing type-mismatch refusal, "This compares X with Y". `SproutList.equals` is deleted; nothing called it but its own spec, and the private sameness that keeps the no-duplicates rule inside a list of lists stays.

## Spec gaps

None.

## Evidence

- `check.spec.ts`: `self.get(:row) == self.get(:row)` and `!=` are refused with the exact words at the left operand's line and column; a list of lists the same; a list against an integer, an object binding or a string gets the generic mismatch and never the list words; `.count == 2` and `.includes(:oak)` still type-check.
- No corpus world: expressions are not yet reachable from a world's files, only through the parser's test entry points, so this refusal has no page to pin until a later item makes them so.

Closes #79

Gate at f2ced4a: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at f2ced4a: `e2e: green (init and check from the installed CLI at f2ced4a)`


ericeslinger:
**Blocking** — `sprout/lang/src/check/check.ts:367-376`

The new list-comparison refusal says "Two lists are not compared with `==`" even when only one operand is a list and the other is not a list at all — the message misdescribes the mismatch.

Verified by evaluating `typeOf` on representative expressions against the `check.spec.ts` fixtures:

- `self.get(:row) == 4` (a `[Ward]` list vs. an integer literal) → "Two lists are not compared with `==`. Ask what a list holds instead: …"
- `self.get(:row) == tool` (a list vs. an object binding) → the same "Two lists" message
- `self.get(:row) == self.get(:note)` (a list vs. a `string` property) → the same "Two lists" message

In each case there is exactly one list involved, so the message asserts something false about what was written, and it fires for every list-vs-non-list mismatch, which is likely more common than list-vs-list. The spec's row is "same type, and not a list": the list-specific message belongs to the case where both operands are the same list type; when the types simply differ, the pre-existing "This compares X with Y" refusal already covers it truthfully. The fix is to take the list branch only when both sides are lists, letting a genuine type mismatch fall through to the generic message.

**Review**: 1 blocking, 0 non-blocking. Gate at f6b728c: green. Small, layering-clean change (delete `SproutList.equals`, add one list guard in `identityType`) whose sole defect is that its new diagnostic's wording is wrong whenever the mismatch is list-vs-non-list rather than list-vs-list. The two-lists cases and the placement before the option-literal and object branches (one refusal, never two) are otherwise correct.

---

ericeslinger:
Fixed in f2ced4a. The list refusal now fires only when both operands are value bindings of the same list type; a list against an integer, an object binding, a string, or a list of another element type falls through to the existing "This compares X with Y" refusal, one per comparison. Tests for the three cases assert the generic words and location and that the list words never appear. Receipts in the body are at f2ced4a.

---

ericeslinger:
**Review**: no blocking findings, no non-blocking findings. Gate at f2ced4a: green. The fix commit narrows the list-comparison guard to fire only when both operands are same-typed value lists (checked via `sameType`), letting every other comparison fall through to the existing generic "This compares X with Y" refusal; verified against the three regression cases, the same-type list cases, and an added `[Ward]` vs `[[Ward]]` case, all producing exactly one correct refusal each. The blocking finding above is resolved.

---

PR #100, opened 2026-09-22
No nesting cap: the parser bounds its own recursion

### LINKED ISSUE #63: `typeExpr` leaves a stray closer when the nesting cap refuses a list type
Not a backlog item. Found by the third `pr-review` pass on #62, while attacking the `skipBracketed` helper that PR added.

## What

`typeExpr()`'s `[` case recurses into itself for the element type, and — unlike `primary()`'s `(` case — never steps over its own bracket when that recursive call fails:

```ts
const element = this.typeExpr();
if (element === null) return null;          // the bracket's closer is left behind
```

`primary()` has the equivalent line written the other way, and the comment on `skipBracketed` says why:

> a bracket the cap refuses is consumed and never opens a level, so its closer is left in the stream — where the loop reading around it takes the closer for its own and ends early, dropping everything after it with nothing said.

That is exactly what still happens here, one level of self-recursion away from where #62's fix reaches.

## The failure

```
:remembers [a: [[[[[[[[[Ward]]]]]]]]] default oak, b: 3]
```

Nine `[`, against a default cap of eight. `parseRemembers` returns a non-null declaration with **no properties at all**, one diagnostic (*"Nothing here may be nested more than 8 deep"*), and **`b: 3` silently gone** — one of the eight orphaned `]` left behind by the outer `typeExpr` frames is consumed as if it were the `:remembers`'s own closing bracket.

## Why it is not urgent

- **Not a regression.** Identical output at #62's head and at its base; the same silent drop already reproduces on `main` before #62 existed, by a slightly different path.
- **Not reachable end to end today.** A message's `carries` type uses the same recursion, but `messageDeclaration`'s `recover()` does not reason about brackets and absorbs the stray closers without losing the next declaration — checked, and the following declaration survives correctly. Only a caller that resynchronises by looking for one specific closing character is exposed, and the one that is (`remembers()`) is not wired into real files yet.

## Why it should be fixed before B12 and B19

Those items add kinds and objects, which is where property bodies — and therefore `:remembers` and written types — are read from real source for the first time. At that point this becomes reachable, and a whole later member vanishing with nothing said is the worst failure this parser has.

## What the fix looks like

Give `typeExpr`'s `[` case the same `skipBracketed(']')` on failure that `primary()`'s `(` case already has, and add the paired spec: a too-deep list type inside a `:remembers` must keep the entry after it, and must not invent a diagnostic about content the author never wrote as one.

While in there, check the other direction too — whether any other construct opens a bracket and abandons it without stepping over it. `skipBracketed` is the helper; the question is whether every site that opens a bracket calls it on every path that gives up.

### LINKED ISSUE #64: A large host `nesting` cap overflows the stack instead of refusing
Not a backlog item. Found by the third `pr-review` pass on #62.

## What

`parse.ts`'s own docblock says *"Problems go to `diagnostics`; nothing is thrown"*, and three specs assert it. With a host-configured `nesting` cap in the low thousands, that stops being true: a deeply nested expression throws an uncaught `RangeError: Maximum call stack size exceeded` out of `Parser.expression` rather than producing a diagnostic.

Reproduced with `caps.nesting: 2999`.

## Why

Recursive descent costs stack per level of real nesting, and `expression()`'s precedence climbing multiplies it: reading one operand descends through every precedence level — `expression(1)` → … → `expression(5)` → `unary` → `postfix` → `primary` — whether or not an operator is written there. That is roughly nine JS frames per bracket, so the cap and V8's stack collide at a cap in the low thousands, long before any performance concern.

This is the same class as the crash the working notes already record for B02 — *"without a bound, seven thousand opening brackets crash the compiler instead of refusing"* — except that the bound is now in place and the cap itself is what exceeds the stack.

## Why it matters, and why it is not urgent

**Limits are the host's.** The spec gives 8 as a default for "expression and block nesting", and a host may set what it likes; nothing says a host's choice may make the compiler crash. A limit whose stated purpose is bounding what a person has to read has no business being set in the thousands, so no sane host will trip this — but "no sane host does that" is not the same as "the compiler answers it with a diagnostic", and the file claims the latter.

Not reachable with the default caps, and nothing in the corpus or the examples goes near it.

## What a fix looks like

Two honest options, and the choice is worth making deliberately rather than by default:

1. **Make the expression reader iterative** — precedence climbing over an explicit operator/operand stack, so depth costs no JS frames. This is the real fix and it also removes the ×9 multiplier. `check.ts`'s `typeOf` already walks its unbounded spines this way, for the same reason, and could be the model.
2. **Bound what a host may ask for**, refusing a `nesting` cap the parser cannot honour, and say so at configuration time rather than mid-parse. Cheaper, but it means the language telling the host a number — which is the one thing Limits says it never does — so it needs a line in the spec if it is chosen.

Either way the acceptance test is the same: a host cap high enough to have crashed produces a refusal, and `nothing is thrown` stays true.

### LINKED ISSUE #80: Remove the nesting cap; the parser bounds its own recursion without a host limit
`nesting` is a host cap in `limits.ts` and the parser counts brackets and prefix signs against it. The spec no longer has a nesting cap. The parser still has to refuse pathologically deep text rather than overflow the stack (this closes #64 and #63 with it); whatever bound it uses is the compiler's own, not a limit the host configures.

Decided by Eric on 2026-09-22 (the review page's spec holes tab) and written into the spec by #72. The code was built before the decision; this brings it in line. Spec section: Limits › Static caps.

ericeslinger:
## What

`nesting` leaves `limits.ts`, the limit table and the bundle's recorded caps, per the spec's Limits ("Nesting has no cap. The compiler bounds its own recursion so that pathologically deep text is refused rather than crashing it; that bound is the compiler's, not a limit the host sets, and it is not recorded in the bundle"). A host passing `nesting` now gets `is not a limit`.

The parser's bound is its own constant, `DEEPEST = 128`, set by the stack rather than by what is readable: reading one bracket costs about nine JS frames, and the costliest shape (parentheses inside call arguments) exhausts Node 22's default stack near depth 700, measured; 128 leaves more than five times over. The refusal names no number: `This is nested too deep to read.` with `Take some of the brackets out.` (or `… signs out.` for a wall of `!`/`-`), once per declaration. This closes #64: a text of any depth produces a diagnostic and nothing is thrown.

This also closes #63: `typeExpr()`'s `[` case now steps over its own bracket on every path that gives up, so a too-deep list type inside a `:remembers` no longer hands its `]` to the enclosing reader and loses the entry after it. The audit of every bracket site found `primary()`, call arguments, `listLiteral`, `remembers()` and the enum reader already correct. One further step was needed for the #63 shape to say one thing: a property whose type could not be read now skips its own `default` and `min`/`max` (`skipPropertyTail`), which were otherwise read as new entries and reported as mistakes the author never made.

## Spec gaps

None. The bound's value is the compiler's affair, and the spec says so.

## Evidence

- `corpus/bad/nested-too-deep` pins the page: a `:remembers` type nested one past the bound, one refusal, and the entry after it kept.
- `parse.spec.ts`: each kind of stack (parentheses, call arguments, prefix signs, list type, list value) parses exactly at the bound and is refused one past it; the refusal names no digit; nothing throws over 50 000 nested expressions or 20 000 nested brackets of each property shape; the #63 pair for a type, a value and a parenthesised expression; the tail skip over every spelling of a default; and the "a defect never loses a well-formed neighbour" tables carry a too-deep type and value, which fail without the fix.
- `limits.spec.ts` and `compile.spec.ts`: `nesting` is not a cap, not in the table, not in a bundle's caps.

Closes #80
Closes #63
Closes #64

Gate at 8298601: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at 8298601: `e2e: green (init and check from the installed CLI at 8298601)`

---


ericeslinger:
**Non-blocking** — `sprout/lang/src/syntax/parse.ts:149` (and `sprout/lang/src/index.ts:16`)

`export const DEEPEST = 128;` is only ever used inside `parse.ts` itself and by `parse.spec.ts`. Because `index.ts` does `export * from './syntax/parse.js';`, `DEEPEST` reaches the published package surface, a number no host or caller has any use for (the spec is explicit that this bound is the parser's own, not a host figure). CLAUDE.md: "Test-only entry points are not exported from `index.ts`." Not new to this PR (`DECLARATIONS` in the same file has the identical pattern), but this doubles it rather than avoiding it.

**Review**: 0 blocking, 1 non-blocking. Gate at 8298601: green (e2e also green). Clean, well-scoped removal of the nesting cap per spec, with a thorough recursion-skip fix for the type and value paths and a generative test that proves no well-formed sibling is lost; the overflow probe reproduced the RangeError near depth 700 for the costliest shape, so the five-times margin holds; no mention of `nesting` survives on any host-facing surface.

---

Reply on the non-blocking finding: leaving it as is. The spec imports `DEEPEST` from `./parse.js`, so removing the `export` keyword would break the spec rather than narrow the surface; keeping the package surface narrower means turning `index.ts`'s `export *` lines into named lists, which is a policy change for the whole package and not this PR's. The constant is also not a test-only entry point: it is the one fact the spec's Limits section says the compiler owns about nesting, and the generated skill (B53) is the reader that would want it, if any.

---

PR #103, opened 2026-09-22
A static cap exceeded refuses at load as it does at publish

### LINKED ISSUE #84: A static cap exceeded at load refuses the world unless the host recorded an exception
`compileBundle` warns at load when a bundle's source or file count exceeds the host's caps. The spec now refuses, unless the host has made an exception for that world; B43 supplies the host's way of recording one, this issue makes load refuse by default.

Decided by Eric on 2026-09-22 (the review page's spec holes tab) and written into the spec by #72. The code was built before the decision; this brings it in line. Spec section: Limits; The compiler › Strict and lenient.

ericeslinger:
## What

The whole-bundle caps, `sourceBytes` and `files`, refuse in both modes. The spec's Limits says a host loading a bundle checked against larger caps than its own refuses to run it unless it has recorded an exception for that world; the host's way of recording one is B43's, so until it lands no exception exists and load refuses as publish does (Eric, 2026-09-22, on the issue). The per-declaration caps were already first-tier refusals and are untouched; `Report.strict` keeps its other two callers.

## Spec gaps

None.

## Evidence

- `compile.spec.ts`: at load, a world past the source cap and one past the file cap are each refused at the manifest's `name` key with no bundle; the publish-mode tests are unchanged.
- No corpus world: both caps default to unbounded, so the corpus cannot exceed them, and the words are unchanged.

Closes #84

Gate at 1d6bc7d: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at 1d6bc7d: `e2e: green (init and check from the installed CLI at 1d6bc7d)`

---


ericeslinger:
**Review**: no blocking findings. Gate at 1d6bc7d: green. Small, well-scoped change (two `strict`→`refuse` calls plus matching spec updates) that matches issue #84's stated scope and the working notes' recorded decision; verified `softenPolicy` cannot soften these refusals (they carry no `since`) and that the distinct `capsExceeding` recorded-caps comparison is legitimately out of scope, unwired and awaiting B43. Considered and not raised: the comment opens by paraphrasing the spec's "checked against larger caps than its own" sentence, which literally describes the recorded-caps comparison, above code that recomputes actual size against the host's current caps; the conclusion it states is right for what is below.

---

Issue #86, opened 2026-09-22
A symbol literal on the left of `==`, and an integer literal outside the operand's range


ericeslinger:
Two findings from the 2026-09-21 review, recorded as open in the notes and not yet decided: the checker refuses `:wet == self.get(:state)` where the spec reads as symmetric, and accepts `self.get(:wear) == 12` on a 0-to-9 property. Eric decides; then this issue implements whichever way.

Decided by Eric on 2026-09-22 (the review page's spec holes tab) and written into the spec by #72. The code was built before the decision; this brings it in line. Spec section: Properties › What the compiler checks.

ericeslinger:
Status after checking with Eric on 2026-09-22: the symbol-literal half is settled as symmetric, since the spec's checker table says "a symbol literal must be one of the operand's options" without naming a side, so `:wet == self.get(:state)` is accepted and checked like `self.get(:state) == :wet`. The integer-literal half (a literal in a comparison outside the other operand's range) is still awaiting a decision; the spec's table refuses an out-of-range literal only for `set` and `remember`.

---
_Generated by [Claude Code](https://claude.ai/code)_

ericeslinger:
The symbol-literal half is done: #105 makes `:wet == self.get(:state)` check exactly as `self.get(:state) == :wet`. This issue stays open for the integer half, which needs Eric's decision: whether `self.get(:wear) == 12` on a property declared `0` to `9` is refused as the compile-time literal case (the false-forever defect the enum rule prevents), or accepted because the spec's checker table refuses an out-of-range literal only for `set` and `remember`. The working notes' _Holes in the spec_ records it as still open.

---
_Generated by [Claude Code](https://claude.ai/code)_

ericeslinger:
Integer literal should fail to compile if we have a priori knowledge that the check will be always false.

PR #105, opened 2026-09-22
A symbol literal on either side of `==` is checked against the other operand

ericeslinger:
## What

`:wet == self.get(:state)` is accepted and checked exactly as `self.get(:state) == :wet` is, per the spec's What the compiler checks ("a symbol literal, on either side, must be one of the other operand's options"), the half of #86 Eric settled on 2026-09-22. `typeOf`'s spine walk stops one node above a `==`/`!=` whose left is a symbol literal instead of handing the literal to `leafType` alone, and `identityType` types the right first and checks the literal against it with the same `option` call the right-hand case uses, so a misspelt option on the left names the options with the same words. Both sides being literals keeps its refusal, which is now reachable.

The second half of #86, an integer literal in a comparison outside the other operand's range, is not decided and is untouched; the issue stays open for it.

## Spec gaps

None new.

## Evidence

- `check.spec.ts`: `:wet == self.get(:state)` and `!=` type as boolean with no refusal; `:slver ==` and `!=` are refused with exactly the right-hand case's words, at the literal; `:wet == 4` gets exactly the words `4 == :wet` gets; `:wet == :dry` still gets "Neither side of this says which enum its option belongs to."
- No corpus world: expressions are not yet reachable from a world's files.

Part of #86

Gate at 95da564: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at 95da564: `e2e: green (init and check from the installed CLI at 95da564)`

---


ericeslinger:
**Non-blocking** — `sprout/lang/src/check/check.ts:384`

`const leftType = left!;` asserts non-null on a parameter typed `BindingType | null`. It is provably safe (the only caller passing `null` is `leafType`'s `'binary'` case, reached exactly when `leftOption` is true, and the function has returned by then) and internal, not a language-boundary read, so CLAUDE.md's typing rule does not forbid it; the file already uses the same pattern for locally provable cases. A discriminated argument would let the compiler check the invariant instead of a comment; worth asking for if the author wants to tighten it, not required.

**Review**: no blocking findings, 1 non-blocking. Gate at 95da564: green. Small, well-scoped fix that makes an already-written "both sides literal" refusal reachable and treats a left-hand symbol literal symmetrically with the right-hand handling; traced nested placements (`&&`, `!`, call arguments) and chained comparisons and found each still typed correctly; integer-range checking in comparisons is untouched, matching the issue staying open.

---

Reply on the non-blocking finding: leaving it as is. The invariant is stated beside the assertion, it is local to one function, and a two-signature shape would add more code than the case earns; if `identityType` grows another caller, that is the moment to discriminate the argument.

---

PR #107, opened 2026-09-22
Refuse a comparison whose integer literal is outside the other operand's range

### LINKED ISSUE #86: A symbol literal on the left of `==`, and an integer literal outside the operand's range

ericeslinger:
## What

A comparison (`==`, `!=`, `<`, `<=`, `>`, `>=`) between an integer literal and an integer whose declared range excludes it is refused at the literal, saying what it always is: `12 is outside 0 to 9, so this is always false.` with the remedy to write a number in the range or take the comparison out. Eric decided this on #86 on 2026-09-22; the spec's checker table carries the rule since #106. One helper serves both the identity and the relation paths and computes the verdict by evaluating the operator against the range's end, which agrees with the other end exactly when the literal is outside. A literal within the range, a full-range integer, two literals, two non-literal operands, and a literal against a non-integer (which keeps its one type-mismatch refusal) are untouched.

## Spec gaps

None.

## Evidence

- `check.spec.ts`: `==`/`!=` with the literal on either side, exact words and location at the literal; a negated literal; both range boundaries accepted; a full-range integer and `.count` against a large literal accepted; two non-literal operands accepted; `"a" == 12` still one refusal; each of the four relations against a literal above and below the range with the computed verdict, in both orientations.
- No corpus world: expressions are not yet reachable from a world's files.

Closes #86

Gate at 03c7e73: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at 03c7e73: `e2e: green (init and check from the installed CLI at 03c7e73)`


ericeslinger:
**Non-blocking** — `sprout/lang/src/check/check.spec.ts:289`

The new test's comment says "`:capacity` is declared `0 min 0 max 9`." but the fixture declares `:capacity 4 min 0 max 9`; the test itself is unaffected since only min and max matter here.

**Review**: 0 blocking, 1 non-blocking. Gate at cc62dd3: green. Small, well-scoped checker addition (one helper plus its two call sites) that correctly refuses an integer literal against `==`, `!=`, `<`, `<=`, `>`, `>=` when it falls outside the other operand's declared range; the verdict evaluated at `min` is sound for all six operators and both literal positions (constant across the interval once the literal is outside), non-literal integer expressions and full-range integers are left alone, a `let` of a ranged property carries the range as the spec's typing table says, and the spec row on the docs branch matches what is implemented.

---

The comment is corrected at 03c7e73, with main (now carrying #106's spec row) merged in; receipts in the body are updated to that head.

---

PR #108, opened 2026-09-22
The world reader reads on past a member it cannot read

### LINKED ISSUE #97: The world reader stops at the first bad property and says nothing about the members after it
Not a backlog item. Found while pinning corpus pages for #77.

## What

A `world` body with a property the parser refuses loses every member after it, with no diagnostic for any of them:

```sprout
world w {
  visitors are Creature
  visitors arrive at hall
  :wear 4 min "x"
  :ward Ward.Iron
  :opens [Ward].brass
}
```

`sprout check` prints one problem, `A min is a whole number.` at line 4, and nothing about lines 5 and 6, which are each a refusal on their own. Swapping the order shows the same: whichever bad property comes first is the only one reported.

## Why it matters

The parser's own rule (its header, and the invariant tests in `parse.spec.ts`) is that a defect in one item never loses a well-formed neighbour in silence, and `:remembers` and enum bodies keep it. The world body does not: an author with three mistakes is told one, fixes it, and is told the next. The file is still refused, so nothing is silently accepted; it is the diagnostics that are lost, which the spec's Diagnostics section is about.

## What a fix looks like

`worldDeclaration()`'s member loop should recover to the next member (the next `:`-symbol or member keyword at the body's own depth) after a member reader returns null, the way `remembers()` resynchronises on `,` and `]`, and the "a defect never loses a well-formed neighbour" invariant should gain a world-body case. Worth doing before B19 reads kind and object bodies with the same loop.

ericeslinger:
## What

A `world` body with a member the parser refuses no longer loses every member after it. After a failed member (or a word that is no member, stepped over once unless a declaration may begin there), `worldDeclaration()` resynchronises to the next member at the body's own depth (a property or `:remembers` symbol, `visitors`, `contains`, or the body's `}`) and reads on; the world is returned with the members that read, since the file's refusal already keeps it from being used and the point is that every member's problem is named (the spec's The compiler › Diagnostics). Braces nest while skipping, and so does a `[` that is closed before any brace, so nothing inside a list default is taken for a member; a `[` never closed is one token and cannot swallow the members after it. The bracket probe is linear: one failed probe records every `[` it walked past, so a stretch of stray brackets is walked once, and the record lives on the parser so successive recoveries do not re-walk it either.

Pinned behaviour that changed, and how: "says what is wrong with one that is not written out" now marks each row lost or kept (name, header and unclosed defects still lose the world; a bad member keeps it with no members); the `#66` invariant over a whole file allows a broken world to survive under its own name; "does not swallow the declaration written after a broken one" keeps its assertions and changes only its comment.

Known limits, not invented around: a bad word directly after a failed member is skipped with it and not named, because members have no separator and recovery can only stop where a member starts (a bad word in first position is named and tested); and an unmatched `{` in the text stepped over desyncs the brace count, so the members after it can be consumed and the world reported as never closed, the same limit `recoverInBraces` has for an enum body.

## Spec gaps

None.

## Evidence

- `corpus/bad/world-two-bad-members` pins the issue's example: three refusals at lines 4, 5 and 6, one page. No other page changed.
- `parse.spec.ts`: the example's three refusals and the surviving `visitors` members; a good property and a `:remembers` after a bad property kept; a list default skipped whole so its `:wet` is not read as a member; an unclosed `[` keeps the member after it, and `[ [:wet]` (an unclosed one followed by a closed run) too; two bad members in a row report both; a world-body case in the `#66` invariant over all six orders of good/defect/good with the anti-vacuity check; and two timing-guarded cases, 100,000 stray `[` after one bad member and 40,000 bad members each ending in a stray `[`, that the quadratic version fails on the default timeout (about 38 s and 22 s measured) and the linear one passes in well under a second. Run against main's parser, the recovery tests fail.

Closes #97

Gate at a619181: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at a619181: `e2e: green (init and check from the installed CLI at a619181)`


ericeslinger:
**Blocking** — `sprout/lang/src/syntax/parse.ts:1131` (`recoverToMember`) and `:1157` (`closedBracketRun`)

`closedBracketRun`'s scan-ahead is not amortized against `recoverToMember`'s own token walk, so a run of stray, individually unclosed `[` tokens at the body's own depth is quadratic in their number: each `[` re-scans forward to the same next `{`/`}`/EOF the previous one already reached. Measured against the built parser with `world w: sprout.World {\n bogus\n [ [ [ … N times … \n}`: 4000 → 73 ms, 8000 → 238 ms, 16000 → 917 ms, 32000 → 3413 ms, ratios converging on 4× per doubling. Nothing upstream bounds this by default (`sourceBytes` is unbounded, and `compileBundle` still parses every file after a cap refusal). Fix: remember that no `]` closes before the next brace once one probe has found so, and skip further probes until a brace is passed, or share a cursor rather than rescanning.

**Non-blocking** — `sprout/lang/src/syntax/parse.ts:1131` (`recoverToMember`)

An unmatched `{` in the tokens stepped over after a failed member desyncs the brace counter, so recovery can consume the real members after it and the world's own `}`, reporting the world as never closed: `:x Zeta silver {` followed by `:y 1` loses `:y` with a refusal that names the wrong thing. It mirrors `recoverInBraces`'s existing limitation for enum bodies and is not silent, but the PR body's known-limit paragraph does not mention it.

**Non-blocking** — `sprout/lang/src/syntax/parse.ts` (whole file, ~1800 lines)

Over the 800-line split threshold CLAUDE.md sets for this file in particular. Pre-existing and out of scope here.

**Review**: 1 blocking, 2 non-blocking. Gate at 8cc2a5c: green. e2e: green. A real, well-tested fix for #97, undone in one respect: new recovery code that is quadratic on pathological input with no default cap to bound it. Checked and found correct: the invariant's coverage of member keywords, types, defaults, bounds, `:remembers` entries and list elements across all six orders; `visitors are 4` then `:x 1` keeps `:x`; a string default holding `{` is safe; the second, true `sprout.World` refusal on a world with a bad member is a correct independent diagnostic; the corpus page pins exactly the three refusals.

---

ericeslinger:
Blocking finding fixed in a619181. Rather than stopping probes until a brace passes (which would misread `[ [:wet]`, where the second `[` does close), one failed probe now records, by source offset, every `[` it walked past, and a later `[` on that record is answered without a scan; the record lives on the parser so a run of bad members each ending in a stray `[` is linear too. Two timing-guarded tests pin it (100,000 stray `[`; 40,000 members each ending in one), both failing on the default timeout against the previous commit. The brace-desync limit is now stated in the body's known-limits paragraph. The file-size note is pre-existing and stays for the split that CLAUDE.md names as its own job.

---

ericeslinger:
**Review**: 0 blocking, 0 non-blocking. Gate at a619181: green. The rescan fix is sound and provably linear, not just in the two pinned shapes: whether a `[` ever closes depends only on the token suffix to the next brace, so a failed probe's record is context-independent, and a closed sub-run inside a failed outer probe is walked at most twice. The two timing-guarded tests pass in 81 ms and 165 ms, the `[ [:wet]` case in 3 ms; the parser's new field sits beside its existing single-purpose recovery state and is not a finding. The blocking finding above is resolved.

---

Issue #111, opened 2026-09-22
A stray `}` in a world body ends the world and loses the members after it in silence


ericeslinger:
Not a backlog item. Found by the generated defect net built for #66, and excluded there by name so the net keeps biting.

## What

A stray `}` inside a world body ends the world, and the file's reader then skips to the next declaration, so every member after the stray brace is lost with nothing said about it. The generator's world-body run has the exact input in its exclusion comment; in shape:

```sprout
world w: sprout.World {
  :alpha 0
  }
  :bravo 1
  visitors are P
}
```

`:bravo` and `visitors are P` vanish.

This is the silent-loss class of #66, and the same limit `recoverToMember` and `recoverInBraces` share: a brace count cannot tell a stray closer from the body's own.

## What a fix looks like

When a world closes and what follows, before the next declaration, is member-shaped (a symbol or a member keyword at the file's own depth), name those members in a diagnostic rather than letting the file reader step past them, as `remembers()` now names entries after a decoy `]`. Then remove the exclusion from the generator.

Issue #110, opened 2026-09-22
A stray `]` inside a nested list default loses the elements after it in silence


ericeslinger:
Not a backlog item. Found by the generated defect net built for #66, and excluded there by name so the net keeps biting.

## What

`:x [[Ward]] default [[oak, silver, ]], [brass, tin]]` loses `brass` and `tin` without naming them: the stray `]` closes the inner list early, the next `]` closes the outer one, and `[brass, tin]]` is read as something after the property rather than as elements of it.

This is the silent-loss class of #66, inside a list of lists.

## What a fix looks like

The list reader, on a stray closer, should look at what follows before taking it as its own end, or name the elements written after the closer, as `remembers()` now names entries after a decoy `]`. Then remove the exclusion from the generator.

Issue #109, opened 2026-09-22
A `:remembers` entry refused before its value is read hands its `]` to the enclosing list


ericeslinger:
Not a backlog item. Found by the generated defect net built for #66, and excluded there by name so the net keeps biting.

## What

When a `:remembers` entry is refused before the rest of it is read (a bad name, a missing `default`, a stray `)`, a type like `%`) and that rest contains a `]`, the rest is read a token at a time and its `]` ends the `:remembers`:

- `:remembers [faulty: ) [[Ward]] default [[oak], [oak]], alpha: 1]` loses `alpha` with nothing said.
- `:remembers [faulty: [Ward] [oak], alpha: 1]` names `alpha`, but blames a `]` the author wrote correctly.

This is the silent-loss class of #66. In the generator's run over `parseRemembers` it is 81 of 1,000 cases; the test asserts the excluded share stays under a fifth so a regression elsewhere is not hidden by it.

## What a fix looks like

Step over a refused entry through its own brackets, the way the world body's `recoverToMember` does after a failed member, so an entry's `]` is never the list's. Then remove the exclusion from the generator and let the run assert the strict rule for these shapes.

PR #112, opened 2026-09-22
A bound is read as a whole number, and the missing-item invariant runs over generated declarations

### LINKED ISSUE #66: The parser's nets have no invariant about a well-formed item going missing

### LINKED ISSUE #67: A bad `min`/`max` bound reports what is wrong inside it, and a signed bracket loses the entry after it
Not a backlog item. It is what #65 tried three times to fix and reverted, and the history is the most useful part of it.

## Two problems, one cause

`propertyBody` reads a `min` or a `max` by reading **any literal** and complaining about its shape afterwards. So whatever went wrong *inside* what was written answers for the bound:

```
:x integer default 0 min [… seventeen things …]
    A list holds at most 16 things.        ← remedy: take some out.
                                              Taking some out still leaves a list
                                              where a number goes.
```

And `literal()` consumes a leading `-` before discovering there are no digits after it, leaving whatever followed the sign behind — which the enclosing loop then walks into a token at a time:

```
:remembers [visits: 0 min -[1], walks: 1]
    properties: []      ← `walks`, written correctly, vanishes with nothing said
```

The second is the worse one: it is the silent-loss class of #66. The first is only a message-quality problem, and it is the one that looks easy.

## Why this is an issue and not a fix

**#65 attempted it three times. Each attempt broke an input beside the one it fixed.**

1. Peek at the bound before reading it; refuse without consuming. → The abandoned construct's closer stayed in the stream, `remembers()` took it for its own, and `:remembers [visits: 0 min [1, 2], walks: 1]` lost `walks`.
2. Refuse, then step over what was written. → Stepping over a bare `]` ate the enclosing `:remembers`'s *real* terminator, and `:remembers [handled: false, visits: 0 min ]` lost `handled` — a property with nothing to do with the bound, and worse than the behaviour it replaced.
3. Stop at any closer instead. → A *decoy* closer then survives: `:remembers [handled: false, visits: 0 min ], walks: 1]` drops `walks`.

Attempts 2 and 3 are the two horns of the same ambiguity, and it is a real one:

> a reader that has not consumed anything cannot tell a stray closing bracket from the one its own caller is waiting for.

Both readings of `min ], walks: 1]` are defensible — the `]` is a typo, or the `]` ends the `:remembers` and the rest is trailing. Resolving it needs the **enclosing closer threaded down to the point of refusal**, which is a change to `propertyBody`'s and `integerBound`'s signatures and to every caller. That is a real design change, not a patch, and doing it at the end of a four-round review is how a fifth round happens.

So #65 reverted to the behaviour on `main`, which is noisier but loses nothing that was not already lost.

## What a fix has to do

- A bound that is not a number says **so**, rather than reporting what is wrong inside what was written — while `min 1.5` keeps *"Sprout has no fractions."*, which is better than the generic sentence.
- `min -[1]` and friends stop losing the entry after them.
- **Neither is bought by losing anything else.** The acceptance test is the invariant in `parse.spec.ts` (`a defect in one item never loses a well-formed neighbour in silence`), extended to cover the two shapes it currently excludes by name.

Likely shape: thread the enclosing closer down so the refusal point knows which bracket is not its to eat. Worth doing alongside #66's wider net rather than before it.

## Not urgent

`property()` is not wired into `file()` yet — kinds and objects are B12 and B19 — so no real `.sprout` file can contain a `min` at all today. `:remembers` is reachable only through its own exported entry point.

ericeslinger:
Two issues in one PR, as #67's own text asks ("worth doing alongside #66's wider net rather than before it"): the net is what proves the bound fix loses nothing else. Three commits: one per issue, and one bounding the new look-ahead after review.

## #67: a bad bound says so, and loses nothing

`min`/`max` are read by `integerBound()`, which takes only a whole number with an optional `-` (the spec's Properties › Declaring a property). Anything else is refused at its first token with `A min is a whole number.`, whatever is wrong inside what was written, and stepped over as one value so the closer after it is left for the loop waiting on it; `min 1.5` keeps `Sprout has no fractions.`. A sign now carries the value it stands before (`-[1]`, `-oak`), in `skipValue` and in `literal()`'s own minus-with-no-number case, which is what was losing the entry after `:remembers [visits: 0 min -[1], walks: 1]`. After a refused bound the property's remaining bounds are skipped as its own text, so `min max 9` is one refusal.

The ambiguity #67's attempts fell on either side of is resolved the narrow way: `min ]` takes nothing, so the `]` ends the `:remembers` and `[handled: false, visits: 0 min ]` keeps `handled`. For the decoy shape `[visits: 0 min ], walks: 1]`, the first `]` is taken as the end and `remembers()` then names every `, name:` entry written after it, in one refusal at the first, stepping through their own `]` where one closes them; the look-ahead stops at a member, a brace, the end of the file, or wherever a declaration may begin, so an unclosed world never lends its neighbour's header to the `:remembers`.

## #66: the generated net

Beside the hand-written tables, a seeded generator (mulberry32, source printed on failure) builds declarations from their parts and injects one defect into any part: a wrong or missing token, a reserved word, a capitalised word, a fraction, a nest too deep, a list over the cap, a bad bound, a bound written twice, a missing or doubled comma, a stray `]`/`}`/`)`, an unclosed `[`; for enums, messages and worlds into names, braces, options, carried types and members; and a world left unclosed with a declaration after it. Five runs, about 3,500 cases: `parseRemembers` (1,000), `parseProperty` whole (400) and list defaults at any depth (600), world bodies (800) and whole files (700) through `parseDeclarations`. Every case holds: a self-contained defect keeps every well-formed item; after a stray closer an item may instead be named; a declaration after an unclosed world is kept or refused in its own text; nothing appears that was not written; nothing throws; the defect refuses (anti-vacuity), and each run fails if a sort of defect, or the unclosed-world class, was never reached. With `parse.ts` put back to main, ten tests fail.

The net found four losses that are not #67's. Each is excluded by name with its exact input and filed: #109 (a `:remembers` entry refused before its value hands its `]` to the list; 81 of 1,000 cases, asserted to stay under a fifth), #110 (a stray `]` inside a nested list default), #111 (a stray `}` in a world body), #114 (a valueless property at the end of an unclosed world takes the next declaration's first word). The two exclusions the tables carried for #67 are removed and covered.

The comment saying nothing reserves an option's name now says what is true: the parser reads `[oak, enum]` as two words and the enum table is what refuses `enum` as an option.

## Spec gaps

None; parser recovery and wording, not language.

## Evidence

- `corpus/bad/bound-not-a-number` pins `A min is a whole number.` at the bracket; no other page changed.
- `parse.spec.ts`: the bound cases above, the removed exclusions now in `ENTRY_DEFECTS`, the unclosed-world repro with both declarations' fates, and the generator. `parse.spec.ts` is about 3,500 lines; `parse.ts` about 1,900, past the split rule, which is #113.

Closes #67
Closes #66

Gate at b11dc83: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at b11dc83: `e2e: green (init and check from the installed CLI at b11dc83)`


ericeslinger:
**Blocking** — `sprout/lang/src/syntax/parse.ts:1422` (`entriesAfterClose`)

`entriesAfterClose`'s look-ahead stops only at `end`, a `symbol`, `{` or `}`. It does not stop where a top-level declaration begins (the way `atRecoveryStop`/`atDeclarationStart` bound the other recovery loops), so when the `:remembers`'s enclosing world is itself unclosed the scan walks past the world's boundary into whatever follows and misattributes tokens there to the `:remembers`. Reproduced against the head:

```
world w: sprout.World {
  :remembers [a: 0]
world bar: sprout.World, name: 1] {
  visitors are P
  visitors arrive at y
}
```

gives `` `name` is written after the `]` that ends this `:remembers`. `` pointing into `bar`'s header (false: nothing was written after `w`'s `]`), then steps through `bar`'s `]`, consuming `world bar: sprout.World, name: 1]` as entries and leaving `{ visitors are P … }` as `w`'s next member (`A world is not made of `{`.`); only `w` survives and `bar` is gone with no diagnostic naming it. Both repros need the following declaration to carry its own malformed `]`, but that is the shape of a second typo made while a brace above is already missing, and the diagnostic is wrong about the source.

It is also a gap in the PR's own net for this class: the generated world-body run always closes its world with a literal `}`, and the whole-file run's one "world never closed" case holds no `:remembers`, so the interaction is unreachable by the generator. Per CLAUDE.md, a resync change without an invariant over generated input for the class it handles is blocking.

Fix direction: bound the look-ahead as `atRecoveryStop` does, stopping at any name that is a declaration keyword, and let the generator produce an unclosed world holding a `:remembers` followed by another declaration.

**Review**: 1 blocking, 0 non-blocking. Gate at 1e2a9b3: green; e2e green. A real, well-tested tightening of bound and value recovery, with one gap in the new look-ahead. Everything else held under attack: `min - max 9` and `min max 9` say one thing; `min 1.5 max 9` keeps the sibling; `skipValue` cannot cross a `}` or `]`; the look-ahead is linear (consecutive `:remembers` bound each other's scan); the three exclusions are narrow and the "under a fifth" guard is meaningful (81 of 773 contained cases); `parse.spec.ts` runs in about 1.5 s against 0.9 s on main; no comment narrates history; the corpus page's column matches the refusal path.

---

ericeslinger:
Fixed in b11dc83. `entriesAfterClose`'s look-ahead now also stops, taking nothing, where `atRecoveryStop` says a declaration may begin (the loose check, since a false stop costs only the naming while a false continuation swallows a declaration's header; the one word spared is an entry's own name after a comma and before a colon, so `], world: 1]` is still named). The repro now yields no false diagnostic: `w` is never closed at `world bar`, and `bar` is refused inside its own header. A direct test pins both declarations' fates and the exact diagnostics; the generator's world-body run now leaves the world unclosed one case in four and follows it with a well-formed or malformed-header declaration, and the whole-file run's never-closed world holds a `:remembers` and is sometimes followed by such a declaration: 52 cases reach the class, and each run fails if it is never reached. The three tests fail against 1e2a9b3's parser. The widening found one further loss outside this bug, excluded by name and filed as #114. Receipts in the body are at b11dc83.

---

ericeslinger:
**Review**: no blocking findings. Gate at b11dc83: green. The fix makes `entriesAfterClose`'s stop reuse the existing loose `atRecoveryStop` question (now offset-aware) with a narrow exemption for an entry legitimately named after a reader keyword; reproduced the original misattribution as fixed, tried `enum`/`message` following declarations (well-formed and malformed), entries named `world`/`enum`/`message` after the `]`, and a normally closed `:remembers` before `message :stir`, all correct; the three new tests fail against 1e2a9b3's parser; the #114 exclusion is scoped exactly to the one still-open bug and hides nothing broader. The blocking finding above is resolved.

---

PR #115, opened 2026-09-22
Split syntax/parse.ts into modules of functions taking the parser

### LINKED ISSUE #113: Split `syntax/parse.ts` into modules of functions taking the parser
Not a backlog item; the split CLAUDE.md's _Where code goes_ calls for. `sprout/lang/src/syntax/parse.ts` is at about 1,900 lines against the rule to split at about 800, and B19 (#19) adds `kind` and `object` declarations to it, so the split lands first.

## What

Split on the file's own section comments into a folder of modules that take a context object, with no behaviour change: the grammar areas (types and literals; properties and `:remembers`; expressions; the declarations: enums, messages, the world; recovery and the depth bound) become modules of functions taking the parser, which keeps its cursor, depth, the reported-once flags and the lexer. `parse.spec.ts` keeps passing unchanged apart from import paths; every test entry point keeps its signature; `index.ts` exports the same names. A file in `syntax/parse/` imports from `source/`, `syntax/` and `bundle/limits` only, as the table allows.

## How to prove no behaviour changed

The gate, the corpus (no page moves), and the `#66` net and its generator, which are what a recovery regression would trip. The commit is mechanical; the PR body says which function went where and nothing else.

ericeslinger:
## What

The split CLAUDE.md's _Where code goes_ calls for, with no behaviour change: not one diagnostic's words or position, not one recovery path, not one exported name or signature. `syntax/parse.ts` (1,909 lines) keeps the public entry points (`parseDeclarations`, `parseProperty`, `parseLet`, `parseExpression`, `parseRemembers`, `DECLARATIONS`, `DEEPEST`) at the same signatures, so `index.ts` and every caller are untouched, and the grammar lives in `syntax/parse/` as functions over the parser context:

| module | holds |
| --- | --- |
| `parse/parser.ts` (252) | the `Parser` context: lexer and cursor, the depth bound and its once flag, the unclosed-bracket record, the readers table, `atDeclarationStart`/`atRecoveryStop`, `deeper`/`reportTooDeep`, `DECLARATION_SHAPES`, `DEEPEST` |
| `parse/recovery.ts` (120) | `skipBracketed`, `separator`, `recover`, `recoverInBraces`, `closedBracketRun` |
| `parse/types.ts` (326) | `typeExpr`, `atType`, `literal`, `listLiteral`, `atFraction`, `skipValue` |
| `parse/properties.ts` (326) | `property`, `propertyBody`, `qualifiedDefault`, `integerBound`, `skipPropertyTail`, `remembers`, `rememberedProperty`, `entriesAfterClose` |
| `parse/world.ts` (288) | `worldDeclaration`, its member readers, `recoverToMember` |
| `parse/declarations.ts` (190) | `file`, `enumDeclaration`, `messageDeclaration`, `DECLARATION_READERS` |
| `parse/expressions.ts` (354) | the precedence table, `expression`, `unary`, `applyPrefix`, `postfix`, `primary`, `argumentList`, `letStatement` |

The one structural change: the readers table is passed into the parser's constructor, so `parser.ts` imports no grammar module and there is no cycle; the fields the modules need are no longer private, and none reaches the package surface, since `index.ts` re-exports `parse.ts` alone.

The spec is carved the same way, each module getting the colocated spec that exercises it, every `describe` moved whole; shared helpers (`read`, `optionsOf`, `shape`, `readProperty`, …) live in `src/fixtures/parse.ts`, which `tsconfig.json` already excludes from the build. No test's text, assertion or expected value changed; the parser's 219 tests are the same before and after (checked by a diff of full test names and states), and the repository's count rises by exactly the 15 boundary tests the new files add. `recovery.spec.ts` is 968 lines because the `#66` invariant is one describe whose helpers call `expect` and so cannot leave a spec file; it stays whole rather than duplicating `nothingVanishes`. The new specs reach their modules through `parse.ts`'s entry points, as the old spec did; the modules are grammar areas of one reader and the entry points are its surface.

Comments changed only where placement made them false ("as the list above", "lives in `parser.ts`", "see the describe above"), and one on the precedence table that still said the spec states no precedence, which it now does.

## Spec gaps

None; a refactor.

## Evidence

- Every corpus page byte-identical; gate and e2e green. The new files' imports stay within `source/`, `syntax/` and `bundle/limits`, checked by hand; no test enforces the layer table.

Left for Eric: CLAUDE.md's _Where code goes_ still names `syntax/parse.ts` as the file most likely to hit the split first; it now has.

Closes #113

Gate at e253a4d: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at e253a4d: `e2e: green (init and check from the installed CLI at e253a4d)`


ericeslinger:
**Non-blocking** — PR body, "Evidence"

The body claims the boundary spec "still pins that a file in `syntax/parse/` imports from `source/`, `syntax/` and `bundle/limits` only". `boundary.spec.ts` rejects only imports that are neither relative nor `zod`/`vitest`/node builtins; nothing in the repository enforces the per-layer table. The new files' import graph does stay within those layers, so the code is fine; the claim about what the test pins is not.

**Non-blocking** — `sprout/lang/src/syntax/parse/*.spec.ts`

None of the seven new colocated specs imports its sibling module's own exported functions; each reaches its module through `parse.ts`'s entry points via the shared fixtures, the same indirection the single old spec used. Read literally, CLAUDE.md's "exercises it directly" is not met. Defensible for a recursive-descent grammar where a type expression is only ever read from inside a declaration; entry-point-level testing is the meaningful level, and direct `Parser` construction would mostly test the harness.

**Review**: 0 blocking, 2 non-blocking. Gate at e253a4d: green. Pure mechanical split; test names (278 per side, 219 in the parser's specs), corpus output and exported signatures are identical before and after by direct measurement, `index.ts` is untouched and re-exports only `parse.ts`, the widened fields do not reach the package surface, and the precedence comment change is accurate against the spec's Precedence section.

---

Replies: the body's evidence line is corrected to say what is true (the import graph was checked by hand; no test enforces the layer table). The second stays as is: the modules are grammar areas of one reader, and its entry points are that reader's surface; a note for Eric, since it is a reading of the house rule rather than of the spec.

---

Issue #19, opened 2026-09-21
B19 · Composition


ericeslinger:
**Backlog item B19** · Phase 3 — Kinds and composition · size L

## What

Composition: the member table, the four rules, the diamond rule, all-run ordering, exclusive collisions refused, property merge by origin with restatement, `without`, composer's own replaces

## Spec

Implements: **Kinds › How members combine, Consent under composition, Suppressing, Properties merge** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#3 (B03)

## Blocks

#20 (B20), #21 (B21), #22 (B22)

## Working on this

1. Do not start until every issue under *Depends on* is closed. Issue numbers equal backlog numbers: item B*nn* is issue #*nn*, and lower numbers come first.
2. Claim it: assign yourself, or comment that you are taking it, before starting.
3. Read the spec sections above and the relevant *Decisions* in the working notes before writing code. Existing code under `sprout/` and `cli/` describes the previous language and is not authoritative; replace it where it disagrees with the spec.
4. Open a PR that references this issue. Where the spec is silent or unclear, say so in the PR and add a line to the working notes' *Holes in the spec* — do not invent quietly.
5. When this closes, add the `ready` label to any issue under *Blocks* whose other dependencies are also closed.

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B19.

ericeslinger:
From #82, decided by Eric on 2026-09-22: `sprout.World` is refused in any kind's or object's composition (the spec's The compiler › What it refuses: "anything but a world composing it"). There is nothing to check until this item parses kinds, so the refusal lands here, with a `corpus/bad` world pinning its words.

---
_Generated by [Claude Code](https://claude.ai/code)_

ericeslinger:
[...]

PR #116, opened 2026-09-22
Kinds and objects: `kind Name: A, b.C { … }` and `object name: A in container { … }` are read

ericeslinger:
The first of three PRs for B19, per the plan on #19: syntax only. Kinds are not resolved here (PR 2), so a kind composing an unknown kind still passes `sprout check`.

## What

`kind Crate: sprout.Container { … }` and `object bench: Bench in composing_room { … }` are parsed into `KindDeclaration` and `ObjectDeclaration` (the spec's Kinds › Declaring and composing; The world model › Objects). A kind may compose any number of kinds, including none, and writes its braces even when empty; an object names its kinds and its container (`in <name>`, as written, for B14 to resolve), and a body may follow, declaring an anonymous kind for it alone. What a world, a kind and an object share, the composition list and the body loop with its recovery to the next member, now lives in `parse/bodies.ts` and serves all three; `parse/kinds.ts` holds the two declarations and `parse/world.ts` keeps only the world's own. Refusals at their token with a remedy: a missing name, missing braces on a kind, a missing `in` or container, an object with no kinds ("An object names its kinds and its container"), and `sprout.World` on a kind or object in the shape tier (`checkKindDeclaration`, per #82).

Two things the widened net found and this PR fixes: a bare `world` or `object` with its name forgotten took the next declaration's first word as its name and its braces as a body; and `object`, being both a type word and a declaration word, made `message :m with` followed by an object on the next line swallow the object, so a type position that meets a declaration start now says the type is missing.

## Spec silences (PR 3 writes them into the notes)

`object` is both a type name and a declaration word; an object with no kinds is refused and one with no body allowed; a kind's braces are required even when empty; an object's own name may be any lower-case word (the reserved-word rule covers options and bindings); whether a bare `World` on a kind stands for `sprout.World` is PR 2's to decide, since the shape tier refuses only the written `sprout.World`.

## Evidence

- `corpus/good/composition` (kinds with none, one and two compositions, a diamond, a restatement, objects with and without a body); `corpus/bad/kind-composes-sprout-world`, `object-no-container`, `object-no-kind` pin their pages. `bad/unknown-word`'s page changes by the words `kind` and `object` joining the list of what the compiler reads.
- `parse/kinds.spec.ts`, `bodies.spec.ts`, `declarations.spec.ts`, `parser.spec.ts`, `types.spec.ts`, `declare/kinds.spec.ts`, `compile.spec.ts`: each refusal and happy shape. The `#66` net runs its member table over world, kind and object bodies (1,206 cases) and its generator over each owner (800 cases each), with kind and object defects in the whole-file tables and generator.

Part of #19

Gate at 06e0fc0: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at 06e0fc0: `e2e: green (init and check from the installed CLI at 06e0fc0)`

---


ericeslinger:
**Non-blocking** — `sprout/lang/src/syntax/parse/recovery.spec.ts` (whole file, 1,069 lines, was 968)

The `#66` net's owner-parameterisation adds real, non-duplicate coverage and each addition is warranted, but the file was already past the split guideline and grows further. Worth splitting into per-concern spec modules in a follow-up.

**Review**: 0 blocking, 1 non-blocking. Gate at 06e0fc0: green. e2e: green. Clean, well-tested syntax-only PR: the world/kind/object body-loop refactor is behaviour-preserving (verified by diffing the moved code against the old `world.ts`, not just by green tests); the two "widened net" bugs are fixed and covered; the `object`-as-type-word ambiguity is handled correctly and tested at exactly the scenarios that could misfire (`message :m with object`, `with` followed by an object declaration, `object` as an enum option, an object with no body before the next declaration); every new refusal names a line and column and says what to write; the one open spec question (an object with no kinds) is flagged in the body for PR 3's notes rather than resolved silently.

---

Reply: the spec split is taken as a follow-up issue rather than in this PR, since the file was over before and the growth here is the coverage the PR is for.

---

PR #118, opened 2026-09-22
Kinds and objects are composed, and properties merge by origin

ericeslinger:
The second of three PRs for B19, per the plan on #19: kinds and objects resolved and composed, properties merged. `resolveWorld` is still not called from `compileBundle` (the standard library cannot travel yet); the world goes through the same composition in PR 3, with `without` and the caps.

## What

- **`declare/kinds.ts`**: `KindRef` gains `order`, the closure in run order with the kind itself last; a `KindTable` on the `EnumTable` pattern (library-qualified identity, two of one name in one library refused at the second, `unqualified` looking in the world then `sprout`).
- **`declare/compose.ts`** (new): one function composes a written list and a body into a `KindRef`, for a kind and for an object's anonymous kind alike (the spec's How members combine, rules 1 to 4 for the members that exist; Properties merge). The closure is depth-first, left to right, each kind at its first appearance, own last: `D: B, C` with `B: A`, `C: A` gives `A, B, C, D`. A cycle is refused once at the `KindExpr` that closes it (`Lidded` composes itself, through `Chest` and `Trunk`); a kind in the loop is said nothing more about. An unknown kind: "Nothing here is a `Wodden`. Did you mean `Wooden`?", a refusal at publish and, at load, a gap under the absent table's `kind-in-composition` row, the objects composing it left out with no second gap. The same kind twice in one list is refused, naming it as written. `sprout.World` composed by a kind or object is refused by identity, which catches an unqualified `World` (a bare name resolves to the world's own kind, else `sprout.X`, whether or not a standard library travelled), skipping what the shape tier already refused. What `sprout.World` is lives in `declare/sprout-world.ts`, read by the shape tier, the composer and the world.
- **Properties**: `ResolvedProperty` gains `origin`. One origin by several paths is one property; two origins are refused at the later `KindExpr` naming both, with the remedy to restate; a restatement in the composer's body becomes the entry with the composer as its origin (so `Crate: sprout.Container { :capacity 40 }` composed beside `sprout.Container` collides on `:capacity` again, and not on `:open`). `restateProperty` in `properties.ts` keeps the type: with no type written, the default is checked against the composed type, so `:ward iron` and `:capacity 40` keep the enum and the range; a written type must be identical (range included); remembered versus plain is a change. `contains`/`contains actors` are OR'd over the closure.
- **`declare/objects.ts`** (new): `ResolvedObject` with its anonymous kind, its container as written (B14 resolves it), and two objects of one name refused only in the same container as written, since "two chests may each hold a `key`"; full scoping is B14's.
- **`bundle/declarations.ts`** (new): the second-tier tables built in dependency order (enums, messages, kinds, objects), out of `compile.ts` (now 679 lines). An `object` in a library's files is refused, as a world is. `Bundle` gains `kinds` and `objects`.

## Spec silences (PR 3 writes them into the notes)

Closure order as post-order; a cycle refused once at its closing edge; the same kind twice refused; the restating kind becomes the origin; a different range, or remembered versus plain, is a change of type; origins that disagree cannot be merged by restating; a bare `World` resolves to `sprout.World` and is refused on a kind or object; an object in a library refused; same-name objects refused only in one container; an object's anonymous kind is named for the object and is the origin of its body's properties; a kind composing an absent kind at load fails without reading its body, and its objects are absent with no gap of their own; the library layout on disk is unrecorded, so the corpus cannot vendor a library and the object-in-library case is pinned in `compile.spec.ts` instead.

## Evidence

- `corpus/good/composition` passes with resolution on, its diamond and restatements resolving as the rules say. `corpus/bad/property-two-origins`, `remembers-two-origins`, `kind-unknown`, `kind-cycle`, `kind-composes-twice`, `restatement-changes-type`, `object-composes-sprout-world` pin their pages; no other page changed.
- `compose.spec.ts`, `kinds.spec.ts`, `sprout-world.spec.ts`, `objects.spec.ts`, `properties.spec.ts`, `types.spec.ts`, `declarations.spec.ts`, `compile.spec.ts`: the order lists (diamond, chain, `E: C, B`, `F: B, A`, an object), which origins are named and where, the type kept by each restatement form, the load-mode gap, the object-in-library refusal in both modes.

Part of #19

Gate at 85dd409: `✓ corpus: every good world passes, every bad one fails as expected`
e2e at 85dd409: `e2e: green (init and check from the installed CLI at 85dd409)`


ericeslinger:
**Blocking** — `sprout/lang/src/declare/compose.ts:268-273`

The "composes twice" refusal displays the resolved fallback identity instead of what the author wrote, unlike the unknown-kind refusal below it, which uses `written`. For a bare, undeclared name written twice, `identityOf` falls back to `sprout.<name>` for lookup, and the `named.has(identity)` branch reuses that guess in the message. Reproduced at 0a17f56 with `kind Crate: Wodden, Wodden { }` and nothing named `Wodden`: `Nothing here is a `Wodden`.` then `` `Crate` composes `sprout.Wodden` twice. `` The second fabricates a standard-library reference that is not on the page. Fix: build the message's name from `written`.

**Non-blocking** — `sprout/lang/src/declare/compose.ts:268-273`

"Composes X twice" is unit-tested but has no `corpus/bad` world pinning its words; the words are right.

**Non-blocking** — `sprout/lang/src/declare/kinds.ts` / `compose.ts`

A two-way runtime import: `kinds.ts` imports `composeKind` from `compose.ts`, and `compose.ts` imports `refuseComposingWorld`, `WORLD`, `writesWorld` from `kinds.ts`. Same layer, no top-level use, no bug today, but real coupling worth breaking by hoisting the world helpers where both can import them.

**Non-blocking** — `sprout/lang/src/bundle/compile.spec.ts`

Already over the split guidance on main (909 lines), now 983.

**Review**: 1 blocking, 3 non-blocking. Gate at 0a17f56: green. e2e: green. Solid, well-tested composition, closure and property-merge implementation with good corpus and spec coverage. Checked against the spec text and found correct: the post-order closure (including `C: B` nesting), the diamond restated on one path collapsing to two origins as disclosed, a restatement in an object's body keeping its type, the unknown-kind message using what was written, one gap per unknown kind at load with the objects of a failed kind absent and nothing else referencing them, typing (every `!` invariant-preserving), layering, and `resolveWorld` still uncalled as the plan says.

---

ericeslinger:
Fixed in 85dd409. The "composes twice" and "Nothing here is a" messages both name the kind as the author wrote it (`writtenKind`: bare stays bare, qualified stays qualified), with spec cases asserting the exact words for both spellings and `corpus/bad/kind-composes-twice` pinning the page. The import cycle is broken by moving `WORLD`, `writesWorld` and `refuseComposingWorld` into `declare/sprout-world.ts` (with its own spec), which the shape tier, the composer and the world all import; `compose.ts` keeps only a type import from `kinds.ts`. `compile.spec.ts`'s size is pre-existing and left. Receipts at 85dd409: gate green, e2e green.

---

ericeslinger:
**Review**: 0 blocking, 0 non-blocking. Gate at 85dd409: green. e2e at 85dd409: green. All three prior findings are confirmed fixed by direct reproduction and by inspecting the compiled module graph: the twice refusal names the kind as written (the change to the unknown-kind message is a pure refactor of the same computation), `corpus/bad/kind-composes-twice` pins the page, and the runtime import graph is now a DAG (`compose.ts` keeps only a type import from `kinds.ts`, erased in `dist`); the new `sprout-world.ts` obeys the house rules with its own spec. The blocking finding above is resolved.

_(Posted on behalf of the `pr-review` agent's second pass. Receipts in the body are updated to 85dd409.)_

---

Issue #14, opened 2026-09-21
B14 · Identifier scope


ericeslinger:
**Backlog item B14** · Phase 2 — The world model · size M

## What

Identifier scope: an identifier belongs to its container and is visible from inside it at any depth, nearest wins; dotted paths for exits; compile-time resolution with run-time range

## Spec

Implements: **Names › Identifiers and scope** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#12 (B12)

## Blocks

Nothing.

## Working on this

1. Do not start until every issue under *Depends on* is closed. Issue numbers equal backlog numbers: item B*nn* is issue #*nn*, and lower numbers come first.
2. Claim it: assign yourself, or comment that you are taking it, before starting.
3. Read the spec sections above and the relevant *Decisions* in the working notes before writing code. Existing code under `sprout/` and `cli/` describes the previous language and is not authoritative; replace it where it disagrees with the spec.
4. Open a PR that references this issue. Where the spec is silent or unclear, say so in the PR and add a line to the working notes' *Holes in the spec* — do not invent quietly.
5. When this closes, add the `ready` label to any issue under *Blocks* whose other dependencies are also closed.

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B14.

ericeslinger:
Taking this one — B13 (#13) is merged, so with B12 closed this is the lowest-numbered `ready` issue with no open dependency.

---
_Generated by [Claude Code](https://claude.ai/code)_

ericeslinger:
Standing down from this — unclaimed and back up for grabs. Nothing was committed or pushed, and the draft is deleted.

Worth recording one thing I ran into before starting, since it is not specific to this issue:

[...]

PR #120, opened 2026-09-22
Identifier scope: objects are placed in one tree, and `in` takes a path

ericeslinger:
Part of #14 (B14 · Identifier scope), the first of two PRs. This one builds the containment tree and resolves every object's `in`. The second resolves `visitors arrive at` and will close the issue. The plan, with the one decision that needs Eric's eye, is on #14.

## What

An object's `in` is now resolved against the declared tree. `declare/tree.ts` builds it: the world is the root, named by the manifest's name, and every object declaration is a node under it, one whose kind is absent included, so what it holds keeps its place for when the kind returns.

**Vantage.** "Nearest wins, seen from the occupant" cannot resolve the occupant's own `in`: with two shelves, `in shelf` is self-consistent from inside either. So what a file writes at its top level is read from inside the world. A first step names something directly in the world, or the world itself, and anything deeper is named by its path: `object key: Key in kiln.shelf`. Nearest-wins is the rule for names inside bodies; `resolveFrom` implements it now, pure and iterative, for exits (B28) and `send` (B32) to call, and its spec pins the chest's own `key`, an occupant naming its container, and a world place reached from any depth.

- **syntax:** an `ObjectPath` node (`syntax/parse/paths.ts`), read for an object's `in` and a world's `visitors arrive at`. A trailing dot, two dots, a number or a capital as a step, and spaces around a dot are each refused at the step, and only that object is lost.
- **declare:** `resolveObjects` only composes; `placeObjects` places in one pass by path length (no recursion, since nesting has no cap) and refuses a ring, an object in itself, a container whose kind holds nothing, two of one name in one *resolved* container, an object named as the world, and the world's name inside a path. A step nothing answers to is a gap under a proposed `container` row of the absent table: refused at publish, the object absent at load, with a did-you-mean and, when the object is deeper, the path to write.
- **bundle:** `Bundle.tree`; `Bundle.objects` holds only objects both composed and placed. The `places` cap now counts every composed object whose kind holds actors, placed or not.
- **corpus:** `good/tree`; `bad/container-{unknown,nested-without-path,cycle,itself,twice-named,holds-nothing,path-malformed,world-in-path}` and `bad/object-named-as-world`. Three existing bad worlds now declare the room their objects sit in, so their pages are unchanged.

## Spec silences, recorded under _Holes in the spec_ (group "Found while building identifier scope (B14)")

- Where `in` is read from: the world, deeper by path. The alternative was a bare name wherever it is unique in the whole tree, which makes a second `shelf` anywhere break a line that did not change.
- Dotted paths in `in` and `visitors arrive at`: the spec shows them only on exits.
- `in` naming something that does not hold things: refused, by analogy with `move`.
- A ring, and an object in itself: not in What it refuses; refused in both modes, the ring once at its first declaration.
- The world's name only as a whole path, and no object may take it.
- Two of one name: refused only in one resolved container; no shadowing warning.
- **Proposed spec change, for Eric:** the absent table has no row for an object whose container is not there. This compiler adds `container`: "the object is absent: not in range, not listed, not addressable; what it holds is unreachable until its container returns". The spec is not edited; the row lives in `bundle/absent.ts` and the notes.
- How the caps count objects that did not place.
- Identifiers in a named kind's body have no vantage (B28, B32); a binding and an object of one name (B23, B32).

## Receipts

- gate at `c8b4be8`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `c8b4be8`: `e2e: green (init and check from the installed CLI at c8b4be8)` (exit 0)

---


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment):

**Review summary: no blocking findings, no non-blocking findings.** Gate re-run at c8b4be8 in a fresh worktree: exit 0, last line `✓ corpus: every good world passes, every bad one fails as expected`.

What it checked: `placeObjects` traced by hand (depth-ordered single pass, including a child written before its container; the single-node and multi-node rings, refused once at the first declaration in source order; cascading silence under a failed container; holds-nothing; two of one name only in one resolved container; the world's name as a whole path only); `resolveFrom` against the spec's three sentences of Identifiers and scope, with the nearest-wins cases at a non-empty vantage and the 3000-deep loop test; every path diagnostic against `corpus/bad/container-path-malformed` and the recovery invariant; the vantage decision and the proposed `container` row recorded as holes rather than invented quietly; layering, comments, colocated specs, and every new corpus page read for correctness. The three existing bad-world fixture edits are necessary and leave their pages unchanged.

---

Issue #129, opened 2026-09-23
Pending wakes per object is a host cap, not a rule of the language


ericeslinger:
Decided by Eric on 2026-09-23 (on the holes-and-decisions doc): "one pending wake per object" cannot be enforced by the compiler, so it is a cap in the host's limits table rather than a rule of the language. A `wake` scheduled past the cap faults, as a `spawn` past the live-instances figure does.

This puts one figure back that #85 took out of `bundle/limits.ts` (`wakesPerObject`). The docs PR that carries this round of decisions moves the line in the spec back into the limits table with a host default; the code follows when B18 (spawn and wake) and B40 (the scheduler) read it. This issue is theirs to close.

ericeslinger:
This is already done on `main`, so I'm closing it.

- The limit on pending wakes per object is the host's `pendingWakesPerObject` runtime budget, in `bundle/limits.ts`. Its default is 1, from the spec's Limits › Runtime budgets.
- A host sets it through `limitsFrom`.
- Asking for a wake past the limit faults with `WakeFault` (`runtime/wakes.ts`) and writes nothing. The visitor is told through the world's `fault` passage, the same way as a `spawn` past the host's bound.
- It landed with B43 (#218) and B36 (#231). Core's `turns.spec` runs it with the limit raised.
- There is no compile-time one-wake rule left.
- The gate is green at 375a19f.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #127, opened 2026-09-23
Design: libraries and extensions, from discovery to vendoring


ericeslinger:
Eric on 2026-09-23: "we need a design for libraries and extensions and how they are discovered, downloaded, hash-checked, and vendored into packages / servers."

Nothing can vendor `sprout` today: the manifest's `libraries` row names a library by name and version, the CLI sends no libraries, and where a library lives on disk is unspecified. Because of that the world is not composed in `compileBundle` (`resolveWorld` is never called), `visitors are` is unchecked there, and every remedy that names a `sprout.*` kind names something that cannot resolve.

The design covers: where a vendored library lives in a world's folder and in a published bundle; how the CLI finds one (a registry, a path, both); the hash the manifest pins and where it is checked; what a server does when the hash it holds matches one it already stores (the notes' future thought: serve the shared copy); extensions alongside libraries; and versioning. Decided alongside: a library's `version` is semver, checked as the world's is once a library can be vendored.

Output: a design page under `docs/design/`, then backlog items for the CLI and the host.

Issue #126, opened 2026-09-23
Composition is written with `is`, not the colon


ericeslinger:
Decided by Eric on 2026-09-23 (on the holes-and-decisions doc): composition moves from the colon to `is`; the colon stays everywhere else as long as it introduces no ambiguity the parser cannot resolve.

```sprout
kind Creature is sprout.Actor, Fragile { … }
kind Marker { }
object brass_key is Key in shelf
world printers_shop is sprout.World { … }
```

Properties (`:season`), messages (`on :fired`), `:remembers` entries and `act` labels (`act nuzzle (target: p)`) keep the colon. A declaration that still writes the colon is refused with a remedy that rewrites it with `is`.

Order: the docs PR that carries this round of decisions rewrites the spec's examples and the Declaring and composing section first; this issue is `ready` once that PR merges. Then: `syntax/parse/bodies.ts` (`composition`), `kinds.ts`, `world.ts` and their specs; the recovery invariant's generator; every `corpus/good` and `corpus/bad` world and its page; `cli/src/init.ts`; the spec fixtures under `sprout/lang/src`. Expect every golden to change; read the diff for pages whose refusal moved rather than merely re-spelled.

ericeslinger:
Claimed, together with #140, on `feat/nested-objects-and-is`.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #125, opened 2026-09-23
Warn when an object's name hides one in a container it can see out to


ericeslinger:
Decided by Eric on 2026-09-23 (on the holes-and-decisions doc): the behaviour of nearest-wins is right, and the compiler adds a warning when a nearer object hides an outer one.

Rule: when an object is placed, and an ancestor container (its container, that container's container, up to the world) holds another object of the same name, warn at the inner declaration: the name now means the inner one from inside its container, and the outer one is reached by its path. Two objects of one name in sibling containers (a `key` in each of two chests) hide nothing and get no warning. Two in one resolved container stay refused.

To do, in `declare/tree.ts` after placement (the tree is complete then, so declaration order does not matter): walk each placement's ancestors' `holds` for its name; emit a warning through the diagnostics with a remedy naming the path of the hidden one. A `corpus/good` world that carries the warning, if the corpus harness records warnings for good worlds (check `scripts/check-corpus.mjs`); otherwise a spec in `tree.spec.ts` pins the words. The spec's list of warnings gains the line in the docs PR that carries this round of decisions.

PR #130, opened 2026-09-23
Spec: the decisions of 2026-09-23

ericeslinger:
Docs only. Writes into the spec every decision Eric made on 2026-09-23 on the holes-and-decisions doc, and the six proposed spec changes he approved there. The working notes move each decided hole out of its "awaiting Eric" group into a "Decided 2026-09-23" block. For Eric to read and merge; no code changes, and the code that now disagrees with the spec has an issue each.

## Decisions written in

- **Range reads the path rule** (The world model › Range): an object reaches a target when nothing strictly between them refuses; a refusing container is reached as a surface from inside and is a wall beyond, in both directions; a bench in a relaying bedroom reaches the house and can `get` its `:season`. Also the order: nearest first, breadth-first a ring at a time, each container's contents in its order; a broadcast delivers in that order. Code: #123, already switched.
- **`without` follows the kind into its composers**; in a diamond, a copy that reaches the composer through a kind that does not leave it out still runs (Declaring and composing). Code: whichever of B22, B24 and B32 first runs a composed member.
- **Arrival at the world is refused** (Actors and visitors; What it refuses). Code: #124.
- **A shadowing warning**: an object hides one of its name held further out, warned at the inner declaration with the path the outer one is now reached by; siblings hide nothing (Identifiers and scope; What it warns about). Code: #125.
- **Composition is written with `is`**: `kind Creature is sprout.Actor, Fragile { … }`, `object brass_key is Key in shelf`, `world printers_shop is sprout.World { … }`, `kind Marker { }`. The colon keeps its other uses (properties, messages, options, `:remembers` entries, `act` labels, a role's or loop variable's kind); a composition with the colon is refused with `is` as the remedy. 36 examples rewritten, the worked microworld included, plus README's. Code: #126.
- **Pending wakes per object is a host cap** (Wakes; Runtime budgets table, default 1; the "two bounds" paragraph; State says "its pending wakes"). Code: #129, for B18 and B40.
- **A library's `version` is semver** (The manifest). Code: with #127.

## The six approved rows

- The absent table gains `a container, in an object's `in``.
- `in` and `visitors arrive at` are read from inside the world, deeper by dotted path without spaces (Objects; Identifiers and scope; the visitor paragraph). Two worked-microworld objects now use paths (`composing_room.paper_store`, `composing_room.cabinet`) because their bare names were nested.
- What it refuses gains: an object inside itself or a ring; an `in` naming something that does not hold things; the world's name as a step or as an object's name; two of one name in one container.
- Static caps: what `objects`, `places` and `kinds` count.
- The range order (above).

## Things I decided while writing that Eric may want to change

- **`is` is not added to the reserved words.** `reserved.spec.ts` checks the code reserves every word the spec lists, so adding it here alone would break the gate; whether it should be reserved is a call for #126.
- **Wording of the colon's remaining uses**, in Declaring and composing: "The colon marks a property, a message or an option in an expression, keys a `:remembers` entry, labels a role in `act`, and gives a role or a loop variable its kind" (`role target: Lockable`, `each p: Creature` keep the colon).
- **Absence's "Catch-up delivers one wake per object"** is untouched; with a cap above 1 it may need a word.
- The notes' "The colon does too much" entry is marked decided for composition only; options spelled `Ward.iron` and `:remembers` entries written as properties stay open.
- Eric's floated alternative, declaring an object inside its container's body instead of `in`, is recorded under Still open, undecided.

Gate at `501039a`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0). No e2e, since nothing under `sprout/` or `cli/` changed except README's example.

---


ericeslinger:
One thing decided after this PR was written, so it is not in it and stays out on purpose: objects will be declared inside the body of what holds them (the parse tree becomes the containment tree, `in` goes; #140). This PR still records that alternative as "not decided" under Still open, and its two worked-microworld objects use dotted `in` paths that #140 will replace. Merging this as it stands is right: #140 gets its own docs PR on top of it, so the `is` rewrite and the nested-object rewrite of every example land as two readable diffs rather than one.

---

ericeslinger:
#145 is stacked on this branch: it carries the nested-object rewrite (#140) and B18's decided lines on top of this PR's spec text. Merge this one first; #145 retargets at `main` when this branch is deleted, and I will verify that afterwards.

---

Issue #16, opened 2026-09-21
B16 · Instance ids and the persisted state model


ericeslinger:
**Backlog item B16** · Phase 2 — The world model · size M

## What

Instance ids and the persisted state model: declared-path ids, minted ids for spawns, container, link destinations, pending wake, per-actor memory, last tick per place, visitor records; a stored value that no longer fits its type falls to the default

## Spec

Implements: **The runtime › State** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#12 (B12)

## Blocks

#17 (B17)

## Working on this

1. Do not start until every issue under *Depends on* is closed. Issue numbers equal backlog numbers: item B*nn* is issue #*nn*, and lower numbers come first.
2. Claim it: assign yourself, or comment that you are taking it, before starting.
3. Read the spec sections above and the relevant *Decisions* in the working notes before writing code. Existing code under `sprout/` and `cli/` describes the previous language and is not authoritative; replace it where it disagrees with the spec.
4. Open a PR that references this issue. Where the spec is silent or unclear, say so in the PR and add a line to the working notes' *Holes in the spec* — do not invent quietly.
5. When this closes, add the `ready` label to any issue under *Blocks* whose other dependencies are also closed.

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B16.

ericeslinger:
Taking this; B14 and B15 are closed. The plan, in two PRs. Four questions for Eric are at the end; none blocks the first PR.

## What B16 is, with no engine yet

[...]

PR #123, opened 2026-09-23
Range: what an object can reach

ericeslinger:
Closes #15 (B15 · Range). The plan is on the issue.

## What

There is no engine yet, no state model (B16) and no pass rules read (B32), so this is the range rule itself: `runtime/range.ts`, a pure iterative walk over a small live-tree interface, with the pass decision and the turn's budget supplied by the caller. B16 implements `LiveTree` over its instance state, B32 evaluates pass rules and uses the same walk for broadcast, B27 resolves nouns against it, B34 turns a spent budget into a fault.

- **The rule** is the one Eric decided on 2026-09-23 (the path rule, in #130's spec text): a target is in range when nothing strictly between asker and target on the containment tree refuses. `rangeOf(context, asker, asking)` walks it: itself; its own contents always (its own rule is never asked: a lid stops others looking in, not the chest looking down), crossing breadth-first into any that passes; then each container outward, reached as a surface and ending the walk if it refuses, else followed by its other contents with the same crossing. So a bench in a relaying bedroom reaches the house (as a surface, so it can `get` its `:season`) and none of the house's other places. Nearest first, breadth-first within each ring, each container's contents in its order; each rule asked at most once. The result carries `reached` with how each came into range (`self`, `held`, `surface`, `passed`), `within`, and `walls` (every container that refused, so an `unreachable` answer can say why). Broadcast is `reached` minus `self` and `surface`, which matches Sending's rules.
- **`reaches(context, asker, target, asking)`**: the same answer by the path between them alone, one step per node climbed, for a `get` or a directed `send` through an identifier.
- **Steps:** one `budget.spend()` per node reached (Limits › Runtime budgets: "every object a range walk visits"); `BudgetExhausted` propagates with no partial result. A walk of 14 nodes costs exactly 14.
- **`asking`** is `any` (for `get`, `each` and nouns) or a `DeclaredMessage`; which of `pass :m` and `pass any` answers is the `PassRule`'s, B32's.
[...]

PR #132, opened 2026-09-23
Instance ids, run-time values and the stored state model

ericeslinger:
Part of #16 (B16 · Instance ids and the persisted state model), the first of two PRs. The plan, with four questions for Eric, is on the issue; none of them blocks this PR. The second PR builds the state itself (the catalogue of a bundle, loading and reloading, the per-turn draft, the live tree for range) on these three modules.

## What

Three modules under `runtime/`, each with its spec, exported from `index.ts`:

- **`ids.ts`**: a branded `InstanceId`. A declared object's id is its declared path with the world first, `printers_shop.kiln.shelf`, the world's own `printers_shop` (The runtime › State); a spawn's is minted, `printers_shop#12`, from the world's one serial counter and never reused. Names match `[a-z][a-z0-9_]*`, so the two forms cannot collide, and both constructors refuse anything that is not a name. `declaredPathOf`, `isMinted`, `idForm` (which form a raw stored string is, for the schema), and a branded `VisitKey`. The brand stops a bare `pathKey` string being passed as an id, pinned with `@ts-expect-error`.
- **`values.ts`**: `Value` (moved here from `lists.ts`, where it was `Element`), `typeKey` (a type's identity as stored state records it: an integer's range is not part of it), `fits` (an integer within its range, an option of its enum, a list within the host's cap whose every element fits), `valueOfLiteral` and `defaultOf`.
- **`stored.ts`**: the persisted JSON form with zod schemas. Each stored property carries the type key it was written under, so a retyped property is never reinterpreted even when the old value would fit; `decodeValue` answers `retyped` or `no-longer-fits`, and the declared default stands ("a stored value that no longer fits its type falls to the default"). Per instance: id, what it was made from (`world`, `declared`, `visitor`, `spawned` with its kind), container, arrival serial, properties, links, pending wakes as a list (bounded by the host's cap, per Eric's decision of 2026-09-23 in #129 and #130), memory keyed by actor id, last tick (null off a place); per visitor: visit key, nickname, instance, last place; the world's `serial`. The schema also checks every id is one of this world's forms, `made` agrees with the id's form, no serial is past the world's, no id or visit key is stored twice, and times are whole seconds. `readStoredWorld` throws `StoredStateUnreadable` (a host defect, not the world's) with zod's message.

Rules pinned in the specs: a stored `"oak"` under a property retyped from string to `Ward` falls to the default even though `oak` is an option; an integer outside a narrowed range is not clamped; a removed option, a list over the host's current cap, a list with one misfit or repeated element, or a value of the wrong shape all default, a list whole; a widened range and an added option keep the value; encode then decode round-trips every value including a nested list and a qualified option; `defaultOf` for every type including `[[Ward]]`.

No corpus change: nothing here is visible to `sprout check`. Nothing reads the clock, filesystem or network.

## Spec silences, recorded under the notes' "Found while building the state model (B16)"

- The minted id format `<world>#<n>`, from one persisted serial shared by spawns, arrivals and wakes, never reused.
- A stored value carries the type key it was written under, so a retyped value is never reinterpreted.
- A value outside a narrowed range falls to the default and is not clamped.
- A list falls to the default whole when any element misfits or it exceeds the host's current cap.
- Times are integer host seconds.
- Pending wakes are stored as a list, bounded by the host's cap (decided 2026-09-23); `lastTick` is on every record and null off a place.

`bundle/limits.ts`'s header still calls one pending wake per object a rule of the language, which matches the spec on `main`; it changes with #129 once #130 merges.

## Review

One blocking finding (wakes as a list against the spec on `main`) argued down on Eric's 2026-09-23 decision, with the comment and notes saying so in 3c04466; one non-blocking (`lastTick` per instance) answered in the same commit.

## Receipts

- gate at `3c04466`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `3c04466`: `e2e: green (init and check from the installed CLI at 3c04466)` (exit 0)


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head 26efb01:

**Blocking** — `sprout/lang/src/runtime/stored.ts` (`StoredInstance.wakes`): pending wakes are modelled and schema-checked as a list with no cap, "because how many may be pending is the host's". The spec on `main` says an object has at most one pending wake, a new request replacing it, "a rule of the language under Time rather than a budget", and State says "its pending wake, if any". As written `readStoredWorld` accepts two or more entries. Suggested: `wake: StoredWake | null`.

**Non-blocking** — `sprout/lang/src/runtime/stored.ts` (`lastTick`): the spec keeps "when it last ticked" per place, not per instance; the field is a harmless always-null on non-places, but the header states it as a per-instance fact.

**Review summary: one blocking, one non-blocking.** Gate at 26efb01: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`. The id formats match the spec's example exactly; `fits`/`decodeValue` never clamp; `retyped` is always distinct from `no-longer-fits` and a retyped value is never reinterpreted; the five recorded holes are genuine and match the code; `readStoredWorld(input: unknown)` is the only `unknown` and the schema is the sole narrowing; nothing reads the clock; the schema's cross-checks stop at structural well-formedness, which is a reasonable boundary; the type-only import cycle between `values.ts` and `lists.ts` is elided at emit and not worth blocking on.

---

**Replies:**

- **Wakes as a list: not a defect, by Eric's decision of 2026-09-23.** On the holes-and-decisions doc Eric said one pending wake per object "feels like a cap since it cannot be enforced by the compiler"; #129 records it and docs PR #130 (assigned to him) writes it into the spec as a row of the host's limits table, default 1, a `wake` past it faulting. The stored form is a list bounded by that cap, which at the default is at most one entry, and PR B's loader reconciles a stored list against the host's figure. B15 (#123) merged the same way, code following a recorded decision ahead of the spec text. 3c04466 says so in the field's comment and adds the line to the notes' B16 group.
- **`lastTick`:** one record shape serves every instance, and the field stays null off a place; the comment now says so (3c04466).

Gate at 3c04466: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`. e2e at 3c04466: `e2e: green (init and check from the installed CLI at 3c04466)`.

---

ericeslinger:
Posted on behalf of the `pr-review` agent: it withdraws the blocking finding on pending wakes stored as a list, on the basis of Eric's 2026-09-23 decision (#129, #130) that the figure is a host cap; the list bounded by that cap is correct for the decided design. Status at 3c04466: no blocking findings; the `lastTick` non-blocking is answered by the comment in 3c04466. Everything else from its first pass stands.

---

Issue #131, opened 2026-09-23
Core records carry the runtime's stored state


ericeslinger:
`sprout/core` still holds the previous state model: `ObjectRecord`, `ActorRecord`, `MemoryRecord` and `nextSpawn`, with `SproutState` an untyped JSON record. Its README and `records.ts` say B16 reshapes them. B16 (#16) does not: reshaping core ripples into `store-sql`'s migrations and `store-document`, and the conformance suite's fixtures, and B16's stored form (`runtime/stored.ts`: `StoredInstance`, `StoredVisitor`, `StoredWorld` with one `serial`) fits core's contract as it stands, since every record is JSON keyed by a non-empty string id.

This issue does the reshaping, after B16 and before B34 (the turn boundary that writes through a store):

- replace the object, actor and memory records with `StoredInstance` and `StoredVisitor` (memory is per instance, keyed by actor id, so `forgetActor` and `exportActor` transpose it);
- move `nextSpawn` onto the stored world's `serial`;
- update the SQL migration, the document store, and the conformance suite's fixtures; the suite itself stays the contract every adapter passes.

Depends on #16. Blocks B34 (#34).

PR #133, opened 2026-09-23
The persisted state model: catalogue, state, loading, the draft and the live tree

### LINKED ISSUE #16: B16 · Instance ids and the persisted state model

ericeslinger:
Closes #16 (B16 · Instance ids and the persisted state model). Second of two PRs, on #132's ids, values and stored form. The plan is on the issue; its four questions for Eric stay open there and are recorded in the notes, none of them changing what this builds.

## What

Five modules under `runtime/`, each with its spec, exported from `index.ts`. The committed `WorldState` is immutable; a write turn works in a `Draft` overlay and commits or drops it, so a fault drops the draft and the world is exactly as it was, a poll reads the committed snapshot, and the overlay is exactly the change set a store transaction needs.

- **`catalogue.ts`**: what one bundle says about instances, built once per load: every placement in `Bundle.tree` with its kind or null and its declared rank; the world's id; `worldKind` and `visitorKind`, both null until a later item puts a composed world and the visitor kind into the bundle (the world is not resolved in `compileBundle` until a `sprout` library can be vendored); the kinds by qualified name for spawns; the arrival id, which is what `Bundle.arrival` becomes; the host's current caps.
- **`state.ts`**: `Instance` (id, what it was made from, its resolved kind, container, arrival serial, a dense property map at the defaults, links, pending wakes, memory keyed by actor, last tick), `VisitorRecord`, `WorldState` (instances, dormant records, visitors, the serial, a derived children index), `StateReader`. Contents order, as B15 handed over: declared objects still where they were declared first, in declared order, then arrivals by arrival serial, so a declared key taken and put back goes last.
- **`load.ts`**: the initial state and reloading are one path, `loadWorld(stored, catalogue)`. A declared placement with a composed kind decodes each stored property against the kind, nothing stored meaning created at the defaults; a declared id with no placement (file withheld or removed, object moved in source, container missing), or a placement whose kind is absent, is dormant and re-saved byte-identical (What absent means: "stored state for absent objects is kept, untouched"); what an absent-kind object holds is instantiated but not live; a spawn of a kind no longer in the bundle is dormant; the world and visitors are dormant while their kinds are unknown; a property no longer declared is dropped, `retyped` and `no-longer-fits` fall to the default, a misfit memory entry is deleted; links, wakes and last tick are kept untouched (B28 reconciles link names, B36 wakes against the cap; B42 applies the arrival rows to a visitor whose place is gone). `saveWorld` writes a canonical order, so save-load-save is the identity and stored order never matters; a store from another world is `StoredStateUnreadable`.
- **`draft.ts`**: `Draft`, the one class (its overlay and counter are state every method needs): read through to the base, `mint` from one serial shared by spawns, arrivals and wakes, `write`, `place` (a new arrival appends to its container's children; the world cannot be placed; nothing may be placed inside itself), `add`, `remove` (only what holds nothing; the world never), `putVisitor`, `destroyed` (readable for the rest of the turn), `commit` once. `storedChanges` turns a commit into the upserts and removals a store transaction takes, leaving out what was added and removed in the same turn.
- **`live.ts`**: `isLive` (decoded, with every container out to the world decoded; the world always; a ring in stored state is simply not live; an iterative climb) and `liveTree(reader)`, the `LiveTree<InstanceId>` range walks.

Specs pin every reload rule one case each over a real world compiled through `compileBundle` (a new spec-only `fixtures/bundle.ts` with a shared shop, loaded whole and with a file withheld); `liveTree` over the initial state agrees with `liveTreeOf(bundle.tree)` on every node including under an absent kind; dropping a draft leaves the base identical; two drafts from one base mint the same ids (core's rule that a body run twice mints the same numbers); destroying `#1` then minting gives `#2`; a reader of the base sees nothing a draft writes; `storedChanges` applied to the base's save equals the new state's save; shuffled stored records give identical loads over eight seeds. `fixtures/live-tree.ts` now builds its ids with `declaredId`, so the two formats cannot drift.

`sprout/core` is untouched but for two comments: its records carry the previous state model until a later item replaces them with the runtime's stored form (#131).

No corpus change. Nothing reads the clock, filesystem or network; times are integer host seconds passed in.

## Spec silences, added to the notes' "Found while building the state model (B16)"

- Contents order: declared objects still where declared, in declared order, then arrivals in arrival order.
- "Its kinds" is stored as what the instance was made from; the closure is re-derived from the bundle.
- A property no longer declared is dropped; plain versus remembered is a change of type.
- A changed default in source does not change existing instances, since the property map is dense.
- A declared object moved at run time into something now absent is unreachable and is not returned to its declared container.
- A visitor instance is retyped to the world's current visitor kind after `visitors are` changes.
- An instance whose container lost `contains` stays where it is.
- Whether dormant instances count toward the host's live-instance limit is B18's.
- Which item puts the composed world and the visitor kind into the bundle is open; until then the world instance has no properties and visitors are dormant.

## Receipts

- gate at `1937618`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `1937618`: `e2e: green (init and check from the installed CLI at 1937618)` (exit 0)

---


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head 1937618:

**Review summary: no blocking findings, no non-blocking findings.** Gate reproduced at 1937618: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`; e2e reproduced: `e2e: green (init and check from the installed CLI at 1937618)`.

What it checked: each of the nine recorded reload holes against `load.ts`, `catalogue.ts` and `state.ts` (each a real gap in State and Absence, none invented quietly, and a moved-in-source object correctly goes dormant as the spec's "kept as for an absent object" says); "kept, untouched" holding through save and load by byte-for-byte equality in the spec; `Draft` earning its class by the overlay state every method reads; `commit` producing a children index consistent with the base and never touching `dormant`; determinism (ids and orderings from rank and serial only; two drafts from one base mint the same ids); `isLive` on a ring, on the world and on an away visitor; the only `unknown` at `loadWorld`'s input, narrowed by the schema; layering, the fixtures kept out of the build and `index.ts`; the core comment edits saying what is true now; every new file's spec asserting on rules; file sizes. One observation, not a finding: the derived `children` index can hold a key for a dormant container whose contents decoded, which nothing reads except through `isLive`.

---

Issue #114, opened 2026-09-22
A property with no value at the end of an unclosed world reads the next declaration's first word as its value


ericeslinger:
Not a backlog item. Found by the generated defect net built for #66 when its world-body run was widened to leave the world unclosed, and excluded there by name so the net keeps biting.

## What

```sprout
world w: sprout.World {
  :faulty :wet
message :omega
```

`:faulty :wet` has no value where one should be, and the world is never closed. The property reader takes `message`, the next declaration's first word, as the value, and `message :omega` is lost with nothing said about it. The world's own "never closed" refusal is still said.

The same reading is what the world-body generator already avoids for a word-led member after a valueless property (`min 0` before `visitors …`); here a whole declaration vanishes.

## What a fix looks like

Where a property's value is missing and the next token is a name, the reader should ask whether a declaration may begin there (`atRecoveryStop`) before taking it as a bare option, and refuse the missing value instead. Then remove the exclusion from the generator so the case is asserted.

ericeslinger:
Work in flight, saved as a WIP commit (adec772) on `fix/valueless-property-before-declaration`: the property reader asks whether a declaration may begin before taking a bare name as a value, with the generator exclusion removed and direct cases started. The builder was stopped before the gate ran, so the tree is untested; resume from that commit or reset the branch to `main` and start over. No PR is open.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #134, opened 2026-09-23
A refused `:remembers` entry is stepped over through its own brackets

### LINKED ISSUE #109: A `:remembers` entry refused before its value is read hands its `]` to the enclosing list

ericeslinger:
Closes #109.

## What

When a `:remembers` entry was refused before its value was read, recovery read what was left of it a token at a time, so a `]` the author wrote correctly inside a nested type or a list default ended the whole `:remembers`, and every entry after it was lost with nothing said. `:remembers [faulty: ) [[Ward]] default [[oak], [oak]], alpha: 1]` lost `alpha`; `:remembers [faulty: [Ward] [oak], alpha: 1]` kept it but blamed a `]` the author wrote correctly.

Now `recoverToEntry` in `syntax/parse/properties.ts` steps over the rest of a refused entry to its own comma or the list's closing `]` at depth zero, by the new `stepPast` in `recovery.ts`, which advances one token or a balanced `[`…`]` run whole through the existing `closedBracketRun` and its `unclosedBrackets` memo; `bodies.ts`'s `recoverToMember` takes the same step, so the two walks cannot drift. No second bracket probe; an unclosed `[` is left where it stands, as everywhere else. Both examples now give exactly one refusal, at `faulty`, and keep `alpha: 1`.

The recovery generator's exclusion for this shape is gone, so its `:remembers` run now asserts the strict rule for every generated shape: 81 of 1,000 cases were excluded before, none after. The exclusions for #110, #111 and #114 stay.

Two hand-written tests had pinned a side effect of the old recovery: `:remembers [a: 0, b c: 1]` used to re-read `c: 1`, the tail of the malformed entry, as a fresh entry and say two things. Under the new rule `c: 1` has no comma before it, so it is the refused entry's own text and is stepped over with the one refusal; the tests now say so.

Corpus: `bad/remembers-entry-recovers` pins the second example. No existing golden changed.

## Review

No blocking findings; one non-blocking (the duplicated walk) fixed in 8adccc1.

## Receipts

- gate at `8adccc1`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `8adccc1`: `e2e: green (init and check from the installed CLI at 8adccc1)` (exit 0)


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head d43c1d3:

**Non-blocking** — `sprout/lang/src/syntax/parse/properties.ts` (`recoverToEntry`): duplicates the walk in `bodies.ts`'s `recoverToMember` almost verbatim (step a token, step a `[` over whole as `closedBracketRun` reports it, respect `atRecoveryStop`), differing only in the stop condition. The shared walk could live in `recovery.ts`, parameterised by what ends it, so the two do not drift.

**Review summary: no blocking findings, one non-blocking.** Gate reproduced at d43c1d3: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`. On the question of `:remembers [a: 0, b c: 1]` no longer yielding `c`: correct, not a regression. Entries are comma-delimited, so `c: 1` is only reachable by re-parsing from wherever the failed read left the cursor, which is what the old code did by accident; the codebase's own "is this a new entry" heuristic (`entriesAfterClose`) requires a comma before the name; `recoverToEntry` stops the instant it reaches a comma, so any properly separated entry survives; and the generator's old fixture model already treated that text as the malformed entry's own tail. The dropped second refusal in `parser.spec.ts` was a double report on one malformed entry's own text. The new corpus page is right in words and position; the 1,000-case run reaches every sort with the exclusion gone; the walk is linear through the memo; an unclosed `:remembers` still yields at a declaration start.

---

**Reply:** the shared step is factored in 8adccc1: `stepPast(p)` in `recovery.ts` advances one token, or the whole `[`…`]` run where the token is a `[` that closes; both `recoverToMember` and `recoverToEntry` take it, keeping only their own stop conditions. Gate at 8adccc1: exit 0, same last line. e2e at 8adccc1: `e2e: green (init and check from the installed CLI at 8adccc1)`.

---

PR #135, opened 2026-09-23
A stray `]` in a list default names what it writes past

### LINKED ISSUE #110: A stray `]` inside a nested list default loses the elements after it in silence

ericeslinger:
Closes #110.

## What

`:x [[Ward]] default [[oak, silver, ]], [brass, tin]]` lost `brass` and `tin` in silence: the stray `]` closed the inner list early, the next `]` closed the outer one, and `[brass, tin]]` was read as something after the property.

Now the outermost list literal, right after it accepts its closing `]`, scans what follows (`elementsAfterClose` in `syntax/parse/types.ts`): every element-shaped token at the list's own depth, options, integers with their sign, strings and `true`/`false`, up to a hard stop (end of file, a symbol, a brace, a `:remembers` entry's own colon-led name, a body member's unpunctuated next word such as `visitors are`, `min`/`max` at the list's depth, or a declaration start). What it finds is refused by name in one diagnostic, "`brass` and `tin` are written after the `]` that ends this list. Everything the list holds goes inside its brackets. Take out the `]` that ends it too early.", and stepped over, together with the one bracket the early close cost the list, so nothing is handed the same mistake twice. A nested list's own close is left alone: what follows it is its enclosing list's next element, indistinguishable at the token level from a stray closer's leftovers, so only the outermost call can judge.

Of the two readings the issue offered, this PR names rather than repairs: a candidate closer's local lookahead cannot tell "closes an inner list, the enclosing one continues" from "stray, the same list continues", and repairing at arbitrary depth would need bracket-repair reasoning; naming keeps every element already read exactly as read and never loses one silently.

The generator's `endsTheListEarly` exclusion is gone, so its list-default run asserts the strict kept-or-named rule for every stray-closer shape: 37 of 600 iterations were excluded before, none after. The generator now draws its element pool from options, integers and strings in turn (seed unchanged), so the rule is asserted for typed lists too. The exclusion for #111 stays.

Corpus: `bad/list-default-stray-closer` pins the issue's example. No existing golden changed. Includes a merge of `main` for #109's `stepPast`.

## Receipts

- gate at `e4be3a1`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `e4be3a1`: `e2e: green (init and check from the installed CLI at e4be3a1)` (exit 0)

---


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head e4be3a1:

**Review summary: no blocking findings, no non-blocking findings.** Gate at e4be3a1: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`. It parsed the false-positive candidates by hand and by running them (`min 0` after a default, a following `:remembers`, a following property, `contains actors`, `without changed`, the end of a body, the next declaration): none misfires, and every stop word the scan relies on is reserved, so no real element can be mistaken for one. A stray closer inside a nested list still propagates to the outermost close, where the scan recovers what followed; the generator's "defect at any depth" run confirms it over 600 iterations with no exclusion. The scan is one forward pass with found tokens consumed at once; timing from 200 to 3,200 properties grew linearly. The corpus page's column 46 is the `b` of `brass`, and its wording parallels the `:remembers` message. No history in the comments; `types.ts` is 437 lines.

---

Issue #17, opened 2026-09-21
B17 · Actors


ericeslinger:
**Backlog item B17** · Phase 2 — The world model · size L

## What

Actors: `sprout.Actor` with its guards (`mover == self`) and `pass any (false)`; the visitor kind; NPCs as objects composing it; the per-world visitor store; per-actor memory keyed by actor id

## Spec

Implements: **The world model › Actors and visitors, What a world may know** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#13 (B13), #16 (B16)

## Blocks

#18 (B18), #22 (B22), #42 (B42)

## Working on this

1. Do not start until every issue under *Depends on* is closed. Issue numbers equal backlog numbers: item B*nn* is issue #*nn*, and lower numbers come first.
2. Claim it: assign yourself, or comment that you are taking it, before starting.
3. Read the spec sections above and the relevant *Decisions* in the working notes before writing code. Existing code under `sprout/` and `cli/` describes the previous language and is not authoritative; replace it where it disagrees with the spec.
4. Open a PR that references this issue. Where the spec is silent or unclear, say so in the PR and add a line to the working notes' *Holes in the spec* — do not invent quietly.
5. When this closes, add the `ready` label to any issue under *Blocks* whose other dependencies are also closed.

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B17.

ericeslinger:
Plan for B17, held until Eric answers the first question below, because it pre-empts part of #127.

## The blocker, and the proposed way through

[...]

Issue #75, opened 2026-09-22
A world's `version` is semver


ericeslinger:
`compileBundle` accepts any non-empty version. The manifest's `version` is now semver; refuse one that is not, at the key, saying what semver looks like.

Decided by Eric on 2026-09-22 (the review page's spec holes tab) and written into the spec by #72. The code was built before the decision; this brings it in line. Spec section: The world model › The manifest.

ericeslinger:
Clarified with Eric on 2026-09-22 before starting: only the manifest's own `version` is checked as semver. Whether a library's `version` (the manifest's `libraries` pins and the version a vendored library declares beside its source) is semver too is not stated, and is recorded under the working notes' _Holes in the spec_ rather than decided here.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #136, opened 2026-09-23
The standard library travels: the CLI carries `sprout`, blessed and pinned

ericeslinger:
Part of #17 (B17 · Actors), the first of two PRs, on Eric's go-ahead there ("Agreed. Build the base stdlib."). The second resolves the world and its visitor kind in `compileBundle` and closes the issue. This one is behaviour-neutral for every world's page: the world is still not composed here.

## What

- **`bundle/standard-library.ts`** exports `STANDARD_LIBRARY`, the `sprout` library (version `0.1.0`, level 1) as real Sprout source, one file per kind, named as the spec's worked microworld writes them: `sprout/world.sprout` (`kind World { contains }`), `sprout/place.sprout` (`kind Place { contains actors }`), `sprout/actor.sprout` (`kind Actor { contains :capacity 8 }`). It holds only what the parser reads today; a Sprout comment in each file names who brings the rest (B20 and B29 the passages, B22 the guards, B32 `pass any (false)`, B48 the verbs). It is a TypeScript module holding Sprout text because the boundary spec forbids `node:fs` in lang; real files plus an embed is #127's refinement. The compiler gives it no privilege: it is vendored and hashed like any library, and a world that pins another hash runs against that copy or none. Its spec pins its hash, that each file reads clean, that it declares exactly the three kinds with their holds and `:capacity`, and that a refusal in a forked copy names `sprout/actor.sprout:2:3`.
- **The CLI carries one copy.** `readWorld` sends it as the vendored `sprout` whenever the manifest names that library, and nothing otherwise, so `compileBundle`'s rule that every named library must travel is unchanged. `checkWorld` blesses its hash ("using the standard library costs an author nothing"); `sprout init` pins it in the manifest it writes. `sprout check`'s `ok:` line now counts the world's own declarations, not the vendored ones. Where a vendored copy of any library lives in a world folder stays unspecified (the notes' line is rewritten to say what is now true).
- **Library semver.** The manifest pin's version and the vendored source's own are checked as the world's is, with parallel wording; the pin-versus-source mismatch warning is skipped when either is not a version, so one problem is said once. Corpus `bad/library-version-not-semver` pins a second library at `"1.0"` (so the pin script never resets it). The notes' hole on this is now marked decided (Eric, 2026-09-23).
- **Every corpus manifest pins `sprout`** (52 of them), written by the new `scripts/pin-standard-library.mjs`, which rewrites only the `sprout` pin and is idempotent; it is the deliberate re-pin step whenever the library changes. The lang fixtures (`fixtures/bundle.ts`, `compile.spec.ts`) vendor the real library where a fake `sprout` stood; the fake in `bundle.spec.ts` stays, since it tests hashing.

Corpus pages: no `expected.txt` changed except `bad/missing-file`, whose one refusal moved from `sprout.json:9` to `sprout.json:15` because the pin, formatted, pushes the `files` key down; the words are the same.

## Notes and things to know

- Working notes: the B19 "Where a library lives on disk" line now says the CLI carries the standard library; the library-semver line is decided; a new B17 group records that every manifest names `sprout`, and that a manifest without it will be refused for a missing `sprout.World` once `compileBundle` composes the world (PR 2).
- `catalogue.spec.ts` now lists `sprout.Actor`, `sprout.Place` and `sprout.World` among spawnable kinds; nothing refuses `spawn World` yet, which is B18's to notice.

## Review

One blocking finding (the stale "not decided" notes line) fixed in 79718d5; no non-blocking.

## Receipts

- gate at `79718d5`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `79718d5`: `e2e: green (init and check from the installed CLI at 79718d5)` (exit 0)


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head f7eed01:

**Blocking** — `docs/design/sprout-working-notes.md:207`: the notes still carry "Whether a library's `version` is semver … Not decided; #75 checks the world's `version` alone", while this PR's `compile.ts` decides exactly that (it checks both the pin's version and the vendored source's own). Code that resolves an open question without saying it is open is blocking, however right the resolution; move the entry to decided with the reasoning.

**Review summary: one blocking, no non-blocking.** Gate at f7eed01: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`; e2e green (`sprout init` gives `ok: 3 declarations in 1 files`, a corpus world checks from the installed CLI). The three kind bodies are exactly what the parser reads today (guards, `pass` and verbs are confirmed unparseable, so the comments' owners are accurate); file names match the worked microworld; the compiler gets no privilege (no new `"sprout"` reference in `compile.ts`; the CLI alone sends and blesses the carried copy, and a pin at another hash goes through the existing mismatch path, both tested); blessing is what Limits › Static caps means; one bad version is said once, and a bad pin for a library that never travelled is two real facts; the pin script touches only the `libraries` block; `ok:` counts own declarations; the kinds-cap figure was recomputed, not absorbed; no `node:fs` in lang; `compile.ts` at 772 lines is under the split line.

---

**Reply:** fixed in 79718d5. The line now reads: "Decided 2026-09-23: it is, and the manifest's pin and the vendored source's own version are checked as the world's is, now that a library can travel." Eric decided it on the holes-and-decisions doc yesterday, and docs PR #130 carries the spec's manifest row; this PR's notes line no longer waits on that merge. Gate at 79718d5: exit 0, same last line.

---

PR #137, opened 2026-09-23
A stray `}` in a body names the members written after it

### LINKED ISSUE #111: A stray `}` in a world body ends the world and loses the members after it in silence

ericeslinger:
Closes #111.

## What

A stray `}` inside a world body ended the world, and the file's reader skipped to the next declaration, so every member after the stray brace was lost with nothing said. Kind and object bodies share the reader and had the same loss.

Now `body()` in `syntax/parse/bodies.ts`, right after it takes a closing `}` (a brace count cannot tell the body's own from a stray one), scans what follows (`membersAfterClose`): brackets are depth-tracked and skipped, a `:remembers` is named entry by entry, any other member-shaped token (a property as written, `:bravo`, or a word the body's readers recognise: `contains`, `visitors`, `without`) is named, and the scan stops at the next declaration start, the end of the file, or the body's own true `}`, which is consumed with the stray run so the file reader never reports it as a stray top-level token. One refusal names everything found: "`:bravo` and `visitors` are written after the `}` that ends `w`. Everything `w` is made of goes inside its braces. Take out the `}` that ends it too early." Nothing is added to the body's members, and a stray `}` followed directly by the next declaration still reads that declaration.

The generator's `endsTheWorldEarly` exclusion is gone, so the world, kind and object body run asserts the strict rule for a stray `}`: 59 of 2,400 generated cases were excluded before (16 world, 24 kind, 19 object), none after; its naming check accepts a property named with its colon. Direct cases in `world.spec.ts` and `bodies.spec.ts`, including a `:remembers` after the brace named entry by entry.

Corpus: `bad/world-stray-brace` pins the issue's example; no existing golden changed. Includes a merge of `main` (#136) and the new world's `sprout` pin.

## Review

No blocking findings; two non-blocking (the dropped colon, a direct test for the `:remembers` path) fixed in 4194721.

## Receipts

- gate at `4194721`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `4194721`: `e2e: green (init and check from the installed CLI at 4194721)` (exit 0)


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head 1586866:

**Non-blocking** — `sprout/lang/src/syntax/parse/bodies.ts` (`membersAfterClose`, the generic branch): a found property is named by its bare text (`bravo`), dropping the colon it was written with, unlike `p.describe()`, `writtenMember()` and this function's own `:remembers` case. An author grepping for `bravo` will not connect it to `:bravo 1` as fast as they would with the colon.

**Non-blocking** — `sprout/lang/src/syntax/parse/bodies.ts` (`remembersEntries`): no direct case in the colocated spec; exercised only through `recovery.spec.ts`'s generated run, which CLAUDE.md says does not count as direct coverage.

**Review summary: no blocking findings, two non-blocking.** Gate at 1586866: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`. No false refusal on correct code: the scan stops only at the declaration readers' words followed by a real start, and body keywords and property symbols are never in that set, so a genuine next declaration is never absorbed (pinned by the "reads the declaration that follows a stray `}` straight away" case). The true close is consumed exactly once and the scan is never re-entered over the same tokens; the lexer's peek buffer keeps it linear. The generated invariant that was disabled for exactly this case now runs over 2,400 cases. Column `8:3` is where `:bravo` starts. No history in the comments; the file is 518 lines. Extending the fix to kind and object bodies is the shared reader's, not scope creep.

---

**Replies (both fixed in 4194721):**

- A property is now named as written, `:bravo`, and the generator's naming check accepts a property with its colon; the corpus page and the two direct cases say `:bravo`.
- `bodies.spec.ts` gains a direct case: a `:remembers [visits: 0, greeted: false]` after the stray brace is named `visits` and `greeted`, entry by entry.

Gate at 4194721: exit 0, same last line. e2e at 4194721: `e2e: green (init and check from the installed CLI at 4194721)`.

---

Issue #117, opened 2026-09-22
Split `parse/recovery.spec.ts` into per-concern spec modules


ericeslinger:
Not a backlog item. `sprout/lang/src/syntax/parse/recovery.spec.ts` is about 1,070 lines against the house rule's split at about 800: it holds the `#66` invariant ("a defect in one item never loses a well-formed neighbour in silence") as one describe, with its defect tables, the seeded generator and the runs over `:remembers`, properties, world, kind and object bodies and whole files. Its helpers call `expect`, so they cannot leave a spec file, which is why it stayed whole through #115 and #116.

## What

Split it by concern (the entry and element tables; the member tables over the three owners; the generated-body run; the whole-file run), with the shared helpers (`nothingVanishes`, `explained`, `reading`, the seeded chooser, the defect tables) in one small spec-support module under `src/fixtures/` that the split files import, the way `fixtures/parse.ts` already serves the parser specs. No test's text, assertion or expected value changes; the count is identical before and after.

PR #143, opened 2026-09-23
A property with no value never takes the next declaration's first word

### LINKED ISSUE #114: A property with no value at the end of an unclosed world reads the next declaration's first word as its value

ericeslinger:
Closes #114, the last of the four parser losses the recovery generator found.

## What

```sprout
world w: sprout.World {
  :faulty :wet
message :omega
```

The property reader took `message`, the next declaration's first word, as a bare option value, and `message :omega` was lost with nothing said. Now, at each of the three places `propertyBody` reads a bare value (no type written, after `default`, and the option after a dot in a qualified default), the reader first asks whether a declaration may begin here; if so it refuses "`:wet` has no value where one should be." with the remedy "Write an option of the type, or a literal, before the next declaration." at the position the value should occupy, and the declaration then reads normally. `skipValue`'s bare-option step (a failed `min`/`max` bound) stops at a declaration start too. A list literal already stopped there. Nowhere else takes a bare word as a value.

The generator's last exclusion is gone with it: the `excluded` bucket and its share assertion are removed from `recovery.spec.ts`, and the value-position and body fuzzers were adjusted so a reserved word that opens a real declaration counts as a reading, not a loss. Direct cases in `properties.spec.ts`, `types.spec.ts` and `world.spec.ts` (the issue's example exactly: the missing-value refusal, `message :omega` read as its own declaration, the never-closed refusal still said).

Corpus: `bad/property-value-before-declaration` pins the page (three refusals: `:wet` is not a value for `:faulty`, `:wet` as a property has no value before the declaration, and the world is never closed). No existing golden changed. Includes merges of `main`.

Resumed from a WIP commit after an interruption; the only thing the WIP lacked was the corpus world.

## Receipts

- gate at `583c773`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `583c773`: `e2e: green (init and check from the installed CLI at 583c773)` (exit 0)

---


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head 583c773:

**Review summary: no blocking findings, no non-blocking findings.** Gate at 583c773: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`; e2e green at the same head. A bare option that equals a declaration keyword cannot exist (all five are reserved and an enum refuses a reserved option name), and the check uses the strict `atDeclarationStart`, so `:a message` with nothing after still reads as an option, as a new case pins. The corpus page's first two refusals are two true facts (`:wet` refused as `:faulty`'s value, then as a property of its own with no value), and the second is what lets `message :omega` survive as a declaration, which `world.spec.ts` proves directly. The generator adjustment stops mis-tallying the newly refused shape without weakening the "nothing vanishes" assertion, and the removed `excluded` assertion was already vacuous on main. The never-closed refusal still stands; comments carry no history; `recovery.spec.ts` at 1,061 lines is pre-existing (#117).

---

Issue #18, opened 2026-09-21
B18 · Spawning and destroying


ericeslinger:
**Backlog item B18** · Phase 2 — The world model · size M

## What

Spawning and destroying: `spawn` as statement and binding, `:spawned`, no consent but `:entered`; `destroy self` at body end with contents falling, bindings readable to turn end, queued messages dropped, place-with-visitor fault; spawn caps per turn and per world per hour

## Spec

Implements: **The world model › Spawning, Destroying** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#17 (B17)

## Blocks

Nothing.

## Working on this

1. Do not start until every issue under *Depends on* is closed. Issue numbers equal backlog numbers: item B*nn* is issue #*nn*, and lower numbers come first.
2. Claim it: assign yourself, or comment that you are taking it, before starting.
3. Read the spec sections above and the relevant *Decisions* in the working notes before writing code. Existing code under `sprout/` and `cli/` describes the previous language and is not authoritative; replace it where it disagrees with the spec.
4. Open a PR that references this issue. Where the spec is silent or unclear, say so in the PR and add a line to the working notes' *Holes in the spec* — do not invent quietly.
5. When this closes, add the `ready` label to any issue under *Blocks* whose other dependencies are also closed.

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B18.

ericeslinger:
From #85, decided by Eric on 2026-09-22: this item's row read "spawn caps per turn and per world per hour". There is no cap on spawns over time any more; the per-turn budget stays, and a `spawn` faults when the host will not hold another live instance, which is the host's storage decision rather than a figure in the limits table. The backlog row is corrected in the docs PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

ericeslinger:
[...]

PR #144, opened 2026-09-23
`spawn` and `destroy` as statements: syntax and the checker

ericeslinger:
Part of #18 (B18 · Spawning and destroying), the first of two PRs. The plan and Eric's decisions are on the issue. The second PR builds the runtime operations over the state model and closes the issue.

## What

- **AST:** `SpawnStatement { spawned: KindExpr; container: ObjectPath }`, `DestroyStatement`, and `LetStatement.value: Expr | SpawnStatement`, since `spawn` is the one statement that yields a binding (the spec's Spawning).
- **Parser:** a new `syntax/parse/statements.ts` reads `spawn K in <path>`, `destroy self` and `let`, which moves here from `expressions.ts`; `parse.ts` offers `parseStatement` in place of `parseLet`. `primary()` refuses `spawn` and `destroy` in expression position, stepping over the same shape `statements.ts` reads. Refusals, each for a non-programmer: `spawn` with no kind, no `in`, `in` without a name; bare `destroy`; `destroy` of anything but `self` ("`destroy` removes only the object whose body runs it. Write `destroy self`. To be rid of something else, send it a message and let it destroy itself."); a word that starts no statement; leftovers after a statement. Where the kind is known, the wording uses it as written.
- **Checker:** a new `check/statements.ts` with `checkLet`, `checkEffect`, `checkSpawn` and `checkDestroy`, taken out of `check.ts` (1,085 → 1,048 lines; its split by expression area is still owed). Refused: a kind nothing declares; `sprout.World`, a bare `World` resolving to it, or a kind composing it ("… is what the world is made of, and there is only ever one world"); a target that is a value ("A new `Cup` goes into something that holds things, and `x` is an integer."); a target of a known kind without `contains`; a target naming nothing in scope; `destroy self` in a body whose `self` composes `sprout.World`. A target of the bare object type (`here`, a handler's `from`) is accepted and checked at run time, since the worked microworld writes `spawn Sheet in here`. No required-property check: every property has a written default.
- **Specs:** beside each new file, with a generated-input invariant over 300 statements dropping every token of each (a well-formed statement never vanishes silently; one dropped token gives exactly one refusal), mutation-checked. The checker's printer's-shop fixture moves to `fixtures/check.ts` for both checker specs; the seeded chooser moves to `fixtures/parse.ts`.
- **No corpus world:** no body can hold a statement yet (handlers and roles are B24's and B32's).

## Spec silences, recorded under the notes' "Found while building spawning and destroying (B18)"

- `spawn` of a kind composing `sprout.World`, and `destroy self` in the world's own body, are refused (decided 2026-09-23, going into What it refuses in the next docs PR).
- A spawn target of the bare object type is accepted at compile and checked at run time.
- An actor kind or the visitor kind may be spawned.
- Which items refuse `spawn` and `destroy` in guards, `permit` and `describe`: B22, B24, B31, which read those bodies.
- The target grammar: `in` takes a binding or an identifier or a dotted path, not an expression.
- A multi-part target is refused as unknown until identifiers inside bodies resolve (B32).

## Review

No blocking findings; two non-blocking (the pre-existing `check.ts` size, owed to its own split; the keep-in-step note on `statementRead`, added in e35808c).

## Receipts

- gate at `e35808c`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `e35808c`: `e2e: green (init and check from the installed CLI at e35808c)` (exit 0)


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head 907b938:

**Non-blocking** — `sprout/lang/src/check/check.ts` (whole file): still 1,048 lines, past the split line; pre-existing, and this PR reduces it.

**Non-blocking** — `sprout/lang/src/syntax/parse/expressions.ts` (`statementRead`): the token-stepping that swallows a `spawn`/`destroy` written where a value is wanted duplicates the shape `statements.ts` reads; it only affects the span of the one refusal, but a form the statement grammar gains later would be stepped over wrongly here. A comment that the two are kept in step, or one shared "how far it extends", would do.

**Review summary: no blocking findings, two non-blocking.** Gate at 907b938: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`. `spawn Kind in <target>` and `destroy self` match Spawning and Destroying exactly; the world refusals are recorded under Holes as decided and going into What it refuses rather than written into the spec here. Accepting a bare-object-type target is required by the worked microworld's `spawn Sheet in here` and is recorded. `checkLet` types a spawn binding exactly as the spawned kind's object type and introduces no binding on a refusal. `fixtures/check.ts` is excluded from the build, unexported, and imports no vitest; the chooser move is a clean dedup. `parseStatement`'s leftover refusal fires only after a statement parsed, on the same call's input. The dropped-token invariant was traced by hand for `in`, a container, a dot and an operator: exactly one refusal each. Layering, comments and exports clean; no corpus world is right while no body can hold a statement.

---

**Replies:**

- `check.ts`'s split by expression area is owed and now named in CLAUDE.md as the next source file to split; not this PR's.
- The keep-in-step comment is added in e35808c: the shape stepped over is the one `statements.ts` reads, and a form the grammar gains is stepped over there too. Gate at e35808c: exit 0, same last line; e2e at e35808c: `e2e: green (init and check from the installed CLI at e35808c)`.

---

Issue #141, opened 2026-09-23
Proposal: `finally destroy self`, cleanup that runs after a body's sends have resolved


ericeslinger:
Recorded from Eric on 2026-09-23 (on the holes-and-decisions doc), deciding B18's question on a destroying body's own sends: they are dropped, so a destroyed object has no effects ("it becomes too messy to debug"). That closes the door on `send x :m` followed by `destroy self` as a one-shot, which a world does want (a match that lights the lamp and is gone).

Eric's sketch: an `andThen` or `finally` form, after Java's cleanup clause, so `send lamp :lit finally destroy self`, where the `finally` fires only after every message the body sent has been delivered and handled. Open questions for the spec: whether `finally` attaches to one statement or to the body; what "resolved" means against the event drain (the end of the cascade the body started, or the end of the turn); whether anything but `destroy self` may follow it; how a fault in the cascade affects it; and what it means for a `finally` inside a handler that was itself reached by a `finally`.

Needs a spec section (Statements; Destroying) before any code. Not scheduled.

ericeslinger:
Eric approved the form on 2026-09-23 ("Agreed. Make it so"); the spec's Destroying now has it, from the evening docs PR: `finally destroy self` stands wherever `destroy self` may and destroys the object at the end of the turn's cascade, once every queued message is handled. The details chosen (end of the whole cascade, only `destroy self` after `finally`, twice is once, a fault drops the mark) are open question 66 in the notes. Code waits on B32 (events and the queue) and B34 (turns and the cascade).

---
_Generated by [Claude Code](https://claude.ai/code)_

ericeslinger:
Built by #224 (B32), following the spec's Destroying section: `finally destroy self` marks the object, and it is destroyed once the queue has drained, in the order marked. Only `destroy self` may follow `finally`, and writing it twice or beside a plain `destroy self` counts as once. The open details are Open 66 (the spec's confirmation) and Open 94 (whether `finally` is reserved). Closing.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #140, opened 2026-09-23
Objects are declared inside the body of what holds them; the parse tree is the containment tree


ericeslinger:
Decided by Eric on 2026-09-23 (on the holes-and-decisions doc): "we should change the parse tree to the original containment tree. Kinds defined outside but containers initial contents defined on the container."

So a container's initial contents are declared inside its body, kinds stay top-level, and `in` goes:

```sprout
kind Chest is sprout.Container { }

world printers_shop is sprout.World {
  visitors are Visitor
  visitors arrive at composing_room

  object composing_room is sprout.Place {
    object cabinet is Chest {
      object brass_key is Key
    }
  }
}
```

Consequences to work through in the spec first (a docs PR, separate from #130), then in code:

- **Spec:** The world model › Objects, Names › Identifiers and scope (the vantage question disappears: an identifier's container is the body it is written in; the dotted path stays for exits and for naming something deeper), the world declaration (the world's body holds its top-level objects), What it refuses (an object outside any body; an object in a kind's body, which has no instance to hold it, unless Eric wants kinds to declare contents for every instance), the worked microworld, and every example that writes `in`.
- **Code:** `syntax/parse/kinds.ts` and `bodies.ts` read `object` as a member of an object's or the world's body; `ObjectDeclaration` loses `container`; `declare/tree.ts` builds the tree from nesting instead of resolving `in` (rings and an object in itself become impossible; two of one name in one body stays refused; the shadowing warning of #125 stays); `bundle/declarations.ts`; `cli/src/init.ts`; every corpus world and `corpus/bad/container-*` (most become moot: unknown container, nested-without-path, cycle, itself, world-in-path); the recovery generator's object shapes.
- **Sequencing:** land with #126 (`is`) so every corpus world is rewritten once. `ready` once the docs PR merges.

ericeslinger:
Claimed, together with #126, on `feat/nested-objects-and-is`, so every corpus world is rewritten once.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #145, opened 2026-09-23
Spec: objects are declared inside what holds them, and B18's decided lines

ericeslinger:
Docs only, stacked on #130 (its base is that PR's branch, so this diff shows only what is new; merge #130 first, and GitHub retargets this at `main` when that branch is deleted). For Eric to read and merge. The decisions are the ones he made on the holes-and-decisions doc on 2026-09-23 after #130 was written.

## Objects are declared inside the body of what holds them (#140)

"We should change the parse tree to the original containment tree. Kinds defined outside but containers initial contents defined on the container." So an `object` is written inside the body of the world or of another object, that body is its container, and `in` goes:

```sprout
kind Chest is sprout.Container { }

world printers_shop is sprout.World {
  visitors are Visitor
  visitors arrive at composing_room

  object composing_room is sprout.Place {
    object cabinet is Chest {
      object brass_key is Key
    }
  }
}
```

- Objects, Identifiers and scope, the world declaration, What it refuses, the absent table, Static caps and A kind body are rewritten: an identifier belongs to the body it is written in and is visible from inside it at any depth, nearest winning; dotted paths stay for exits and for naming something deeper from outside; a kind's body never holds an object (a kind has no instance to hold it); the world's body is one block in one file, so every object in a world is declared in the file that declares the world, while kinds, enums, verbs and messages may sit in any file. The refusals for an object inside itself, a ring, an unknown container, and the world's name as an object's name go, with the absent table's `container` row; an object inside something whose kind does not hold things, two of one name in one body, and the shadowing warning stay.
- 23 `object … in …` declarations in 11 fenced blocks are rewritten to the nested form, the worked microworld included: its objects now sit in the world's body in `world.sprout`, and its kinds move to `locks.sprout` (Warded, Key) and `press.sprout` (Rib, Sheet). Excerpts that show an object alone are covered by one sentence under Objects.
- One new paragraph under What it surfaced, "The world file is the floor plan"; cut it if unwanted.

## B18's lines

- What it refuses gains `spawn` of `sprout.World` or a kind composing it, and `destroy self` in the world's own body.
- Destroying: a destroyed object has no effects; everything pending on it is dropped: messages queued to it, messages it sent not yet delivered, engine messages naming it as their `from`, and its pending wakes. Contents still fall to its container as written, and "a visitor standing directly in it" is the fault (Eric's preference for destroying contents instead is recorded under Still open, pending the visitor rule that goes with it).
- Spawning: a target out of range or that does not hold things at run time is a fault, never nothing; a spawn into the world is allowed wherever the world is in range.
- The absent table gains "a kind, in a `spawn`": the `spawn` faults when it runs.
- The host's live-instance bound counts every instance it stores, dormant ones included.

## Holes added under Still open

- Whether a world's body may be continued across files, and what lenient loading does with a world file that fails to compile (every object is in it, so the world would admit no one).
- Whether a kind may declare initial contents for every instance (not now).
- Contents of a destroyed object: fall or destroyed with it (Eric prefers destroyed, pending the visitor rule).
- `finally destroy self`, #141.
- **A contradiction left in the spec on purpose:** Destroying still says the container receives `:entered` for each thing that fell, and that message's `from` is the destroyed object, which the new rule drops. Both sentences stand until Eric says which wins; the notes record it.

The B18 group's decided line is dropped from the notes, and its target-grammar line no longer compares to `in`.

`docs/` is prettier-ignored; the two files pass prettier when checked directly. No code.

---


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head aadc306:

**Review summary: no blocking findings, no non-blocking findings.** Docs only, no gate. Every `object … in …` example is rewritten or covered by the excerpt sentence, with no dangling reference to a container "as written" or to an object's `in` anywhere in the spec. Objects, the world-model intro, Range, Places inside places, Identifiers and scope (shadowing and dotted paths), Addressing and display, Static caps, the absent table and What it refuses read consistently with the containment-tree decision. The worked microworld is complete after the file moves: every object nested in `world.sprout`, kinds only in `locks.sprout` and `press.sprout`, the prose files unchanged in content. The admitted contradiction (fallen contents' `:entered` naming the destroyed object as `from`) and Eric's preference for destroying contents both stand under Still open, not resolved in the spec; the Decided block lists each decision once and the superseded awaiting lines are gone. The new paragraphs read as rules in the spec's voice.

---

ericeslinger:
Added at 2a8a792: the twelve holes found building passages (B20, PRs #150 and #154), as a group under _Holes in the spec_ after the B18 one. The ones that need a decision rather than a nod: a composer's own `default` beside a composed non-default (built: the composer's own applies), a default rewritten over a default (built: collides), and that a register library replaces stock lines only by writing non-defaults, since two defaults collide.

---

Issue #32, opened 2026-09-21
B32 · Events


ericeslinger:
**Backlog item B32** · Phase 4 — Verbs, grammar and prose · size M

## What

Events: `send` to an identifier or binding, the broadcast algorithm, `pass :m` and `pass any`, breadth-first drain after the effect pass, `changed` hooks queued once per change, event and cascade budgets, the unsent/unhandled warnings

## Spec

Implements: **Events, messages and the bus** in [`docs/design/sprout-design-spec.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-design-spec.md). The spec is the end state; where this issue and the spec disagree, the spec wins and this issue is wrong. The reasoning behind the design is in [`sprout-working-notes.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-working-notes.md) (see *Decisions*), and the reviews that shaped it are in [`docs/design/2026-09-20-reviews/`](https://github.com/overstory-social/sprout/tree/main/docs/design/2026-09-20-reviews).

## Depends on

#15 (B15), #7 (B07)

## Blocks

#34 (B34)

## Working on this

1. Do not start until every issue under *Depends on* is closed. Issue numbers equal backlog numbers: item B*nn* is issue #*nn*, and lower numbers come first.
2. Claim it: assign yourself, or comment that you are taking it, before starting.
3. Read the spec sections above and the relevant *Decisions* in the working notes before writing code. Existing code under `sprout/` and `cli/` describes the previous language and is not authoritative; replace it where it disagrees with the spec.
4. Open a PR that references this issue. Where the spec is silent or unclear, say so in the PR and add a line to the working notes' *Holes in the spec* — do not invent quietly.
5. When this closes, add the `ready` label to any issue under *Blocks* whose other dependencies are also closed.

Source: [`docs/design/sprout-build-backlog.md`](https://github.com/overstory-social/sprout/blob/main/docs/design/sprout-build-backlog.md), row B32.

ericeslinger:
[...]

PR #146, opened 2026-09-23
Spawning and destroying at run time: `spawnInstance` and `destroyInstance` over the draft

### LINKED ISSUE #18: B18 · Spawning and destroying

ericeslinger:
Closes #18 (B18 · Spawning and destroying). Second of two PRs, on #144's syntax and checker. The plan and Eric's decisions of 2026-09-23 are on the issue.

## What

`runtime/lifecycle.ts`, pure functions over the state model's `Draft`, exported from `index.ts`. Calling them at a `spawn` statement and at a body's end is B24's and B32's; the queue that drops and delivers their sends is B32's; the turn boundary that turns a fault into the `fault` passage is B34's.

- **`spawnInstance(context, spawner, kind, container)`**, writing nothing until every check passes, in order: the kind is one a spawn may name (else the fault "`Cup` is absent, so it could not be spawned."); the container is in range of the spawner by `reaches` (charged to steps) and live (else "`<container>` is out of range of `<spawner>`, so nothing could be spawned in it."); the container holds things, the world always (else "`<container>` holds nothing, so `Cup` could not be spawned in it."); `budget.spawn()` past the host's spawns per turn throws as before; the host's live-instance bound, `mayHold`, a per-turn number recorded with the turn rather than a callback so replay reproduces it, counted against everything stored, dormant included ("the host will hold no more instances in this world, so `Cup` could not be spawned."). Then it mints an id and adds the instance at its defaults, last in its container, and returns the engine sends in order: `:entered` to the container, `:spawned` to the new object, both `from` the spawner. A spawn into the world is allowed where the world is in range.
- **`destroyInstance(draft, id)`**: the world cannot be destroyed; a visitor's instance cannot ("a person is never destroyed"); a place with a visitor standing directly in it cannot ("destroying it would move them without a word"). Otherwise its contents fall to its container in their held order, each landing last, and it is removed, the record readable to turn end through `draft.destroyed`. It returns one `:entered` per fallen thing, `from` the destroyed object; under Eric's no-effects rule the queue drops sends naming a destroyed `from`, so the rule lives in one place, the queue's, and the notes record that the spec still carries both sentences (a reminder is on #32). Dormant contents are left where they are (their records are kept untouched; recorded as a hole).
- **Faults** are a `LifecycleFault` class thrown like `ListFull`, each with its reason and text; `reason` and `object` are what the `fault` passage renders by noun, the detail names ids for the log and the host. A destroy of an id that is no instance is an engine defect, a plain `Error`.
- **`Draft.held`** counts what is stored (decoded and dormant), up one on `add`, down one on `remove`, so spawn-then-destroy in one turn nets to zero. **`catalogue.kinds`** leaves out kinds composing `sprout.World`, so a stored spawn of it stays dormant (pinned in `load.spec.ts`). `budget.spawn()`'s comment says the cap is per turn only.
- **Specs:** every case above with its text and with nothing written when a fault fires; a spawn into the spawner itself; a visitor deeper inside that falls with what holds them; an NPC falling like anything else; an id never reused after a destroy; and a generated invariant over 60 random spawn, destroy and place steps for each of 8 seeds: every instance stays live, `children` agrees with `container`, `held` matches the stored count, and a fault writes nothing.
- No corpus change: nothing here is visible to `sprout check`.

## Working notes

The B18 group records the nine decisions (contents fall as written, with Eric's preference for destroying them pending the visitor rule), dormant contents left in place, the `:entered`/no-effects contradiction the spec still carries, the host's bound as a recorded number, the falling order and the `from` of `:entered`, `destroy self` twice being once (the body's rule), and memory keyed by a destroyed NPC kept. B16's dormant-count line is closed.

Not in this PR: the absent-table row "a kind, in a `spawn`" is decided and written into the spec by #145, but `bundle/absent.ts` gains it only when something records an absence of that kind, which nothing does until a body can hold a `spawn`.

## Review

No blocking findings; two non-blocking (fault text as data for B34, noted in 2e86cf3; the `:entered` tension, left for Eric with a reminder on #32).

## Receipts

- gate at `2e86cf3`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `2e86cf3`: `e2e: green (init and check from the installed CLI at 2e86cf3)` (exit 0)


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head a7910ab:

**Non-blocking** — `sprout/lang/src/runtime/lifecycle.ts` (the fault texts): `LifecycleFault`'s detail embeds raw ids, dotted declared paths and the minted `printers_shop#1`, which no author writes and which read like a stack trace next to the plain-language refusals elsewhere; `shownKind` shows the file already cares for kinds. Likely structured data pending B34's rendering rather than final player text; a note saying so would head off the question.

**Non-blocking** — `lifecycle.ts` (`destroyInstance`'s sends) and the notes' B18 line: the `:entered` per fallen item carries `from: id`, and the queue B32 builds drops sends naming a destroyed `from`, so the container is never told, against Destroying's "the container receives `:entered` for each thing that fell". Recorded plainly in the code and the notes, so not a quiet resolution; but this PR bakes that shape into data a later PR must special-case, or the contradiction resolves by default the way the spec's words say is wrong. A reminder on B32's issue, not a fix here.

**Review summary: no blocking findings, two non-blocking.** Gate at a7910ab: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`. `budget.spawn()` before the `mayHold` check spends nothing that survives, since a fault abandons the transaction and `faultsWritingNothing` pins it; `instance.container!` is sound because the only null container is an away visitor, refused earlier; `Draft.held` moves only on `add` and `remove` and the eight-seed fuzz holds it true; `catalogue.kinds` correctly leaves out only kinds composing `sprout.World`; layering, sizes, determinism and comments clean.

---

**Replies:**

- The doc comment on `LifecycleFault` now says which is which (2e86cf3): `reason` and `object` are what the `fault` passage renders, naming the object by its noun; the detail names it by id for the log and the host, never for a visitor. Gate at 2e86cf3: exit 0, same last line; e2e at 2e86cf3: `e2e: green (init and check from the installed CLI at 2e86cf3)`.
- The `:entered` tension is left for Eric to decide (it is under Still open in #145 and on the holes doc) and a reminder is posted on #32 so B32 does not resolve it by default.

---

PR #147, opened 2026-09-23
A library that did not travel is refused once

### LINKED ISSUE #139: A library that did not travel is refused once, not once per kind it would have supplied
Since every world composes `sprout.World` and a visitor kind composes `sprout.Actor` (#136, #138), a `sprout` pin at a wrong hash, or a manifest that names a library whose source did not travel, is refused at publish four times: the mismatch (or absence) at the manifest, then `sprout.World`, `sprout.Actor` and `sprout.Place` each reported as a kind nothing declares. The three kind refusals are consequences of the first, and a non-programmer reads them as three more problems.

To do: when a library the manifest names is absent or mismatched, the references into it are not refused again at publish; the manifest's refusal is all that is said (at load they stay gaps, as the absent table's rows say). The same pattern already exists for the missing world (no arrival or visitor gap while an own file was first-tier refused). Pin the single-refusal page in `corpus/bad/library-mismatched` and update `cli/src/check.spec.ts`, which today expects all four.

Not a backlog item; found while building B17.

ericeslinger:
Closes #139.

## What

Since every world composes `sprout.World` and a visitor kind composes `sprout.Actor`, a `sprout` pin at a wrong hash, or a manifest naming a library whose source did not travel, was refused at publish four times: the mismatch or absence at the manifest, then `sprout.World`, `sprout.Actor` and `sprout.Place` each as a kind nothing declares. The three were consequences of the first.

Now the manifest's refusal is all that is said. `Report` (in `bundle/compile/report.ts`) keeps the names of the libraries `checkLibraries` marked unusable, missing or mismatched, and at publish `gap()` says nothing for a `kind-in-composition` or `world` gap whose written kind is qualified with one of them. That one place covers every emission site, since the kind table, the objects and `composeWorld` all report through `report.gap`. A kind written without a library that resolves to nothing is still refused as before, and load mode is untouched: an absent library's kinds stay `kind-in-composition` gaps, as the absent table says.

`corpus/bad/library-mismatched` pins the one refusal ("The library "sprout" that travelled is not the source the manifest recorded."); `cli/src/check.spec.ts` expects one. `corpus/bad/library-sprout-unnamed` keeps its two refusals on purpose: its manifest never names `sprout`, so no library is absent or mismatched and there is no manifest refusal for the kind refusals to be consequences of; Libraries and namespaces defines absence for a library the manifest names. No other golden changed.

## Receipts

- gate at `b0e68a6`: `✓ corpus: every good world passes, every bad one fails as expected` (exit 0)
- e2e at `b0e68a6`: `e2e: green (init and check from the installed CLI at b0e68a6)` (exit 0)

---


ericeslinger:
Posted on behalf of the `pr-review` agent (it cannot post from its environment), on head b0e68a6:

**Review summary: no blocking findings, no non-blocking findings.** Gate at b0e68a6: exit 0, `✓ corpus: every good world passes, every bad one fails as expected`; e2e green at the same head. Every `report.gap()` call site was traced: the suppression fires at publish only, for `kind-in-composition` and `world` gaps whose written kind carries a library prefix `checkLibraries` marked unusable, and `report.spec.ts`'s five new cases assert the rule directly (suppressed with a matching prefix; a bare name still refused; another library's gap still refused; nothing changes at load; a dotted `container` row untouched). At publish the bundle is null whenever the manifest refusal exists, so `absent` is dead there anyway. The visitor kind's "not an actor" refusal goes through `diagnostics.refuse`, not the gap, and still fires; an unknown visitor kind from a mismatched library is a consequence and is rightly suppressed. `library-sprout-unnamed` keeping two refusals is a real distinction: an unpinned library is never marked refused, so there is no manifest refusal for the kind gaps to follow from. One thing worth knowing, not changing: `library.Nonesuch`, a kind the real library would not have either, is masked the same way once the library is unusable; nothing verifiable is lost, since the manifest refusal already blocks publish. `Report`'s new bookkeeping is the same shape as its accumulators, no new god object; comments cite the spec with no history. The change extends the notes' existing rule that a consequence of one root problem is said once, so it needs no new hole.

---

PR #148, opened 2026-09-23
Split the parser recovery spec by concern

### LINKED ISSUE #117: Split `parse/recovery.spec.ts` into per-concern spec modules

ericeslinger:
## What

`syntax/parse/recovery.spec.ts` had grown past the line CLAUDE.md draws, and most of it exercised the parser as a whole rather than `recovery.ts`. This PR splits it on its section comments:

- `syntax/parse/recovery.spec.ts` is rewritten to exercise `recovery.ts` directly: `separator`, `recover`, `recoverInBraces`, `closedBracketRun`, `stepPast` and `skipBracketed` each get their own cases (21 new direct cases).
- `syntax/parse/recovery/{neighbours,remembers,properties,bodies,file}.spec.ts` hold the parser-level recovery specs, one concern each: the neighbour invariants, `remembers` entries, property values, kind bodies, and whole-file resync.
- The shared machinery (the generated-world builder, the `chooser`, the finding collectors) moves to `src/fixtures/recovery.ts`. It returns finding arrays instead of calling `expect`, because the boundary spec forbids vitest outside spec files; `fixtures/` is excluded from the package build.

No parser behaviour changes. Tests go from 324 to 345.

Closes #117

## Receipts

Gate at `3e4fe7b`:

```
✓ corpus: every good world passes, every bad one fails as expected
```

e2e at `3e4fe7b`:

```
e2e: green (init and check from the installed CLI at 3e4fe7b)
```

---


ericeslinger:
Posted on behalf of the pr-review agent (it has no GitHub access in this session):

**Review**: no blocking findings. Gate at 3e4fe7b: green (exit 0, last line "✓ corpus: every good world passes, every bad one fails as expected"). Faithful, test-only split of an over-length spec file per #117: every prior case/seed/count is preserved, plus new direct coverage of `recovery.ts`'s own exports that main lacked; the shared generator machinery moved to `fixtures/recovery.ts`, correctly excluded from the build and not exported.

No per-finding comments: it compared every defect table, seed, iteration count and assertion in the old `recovery.spec.ts` against the split files and found nothing dropped or weakened. Merging.

---

